PR #14, opened 2021-12-17
Properly ssz hash the transactions list 

jparyani:
Also be more permissive on snake_case/camelCase for blinded blocks.


go.mod
```diff
@@ -13,5 +14,6 @@ require (
 
 replace (
 	github.com/ethereum/go-ethereum => github.com/MariusVanDerWijden/go-ethereum v1.8.22-0.20211208130742-dd90624af970
+	github.com/ferranbt/fastssz => github.com/kasey/fastssz v0.0.0-20211117233602-01b170cf6dcc
```
terencechain, line 17: The latest `prysmaticlabs/fssz` is better. The latest fixes and upcoming optimizations are happening there:
https://github.com/prysmaticlabs/fastssz

lib/service.go
```diff
@@ -225,15 +236,25 @@ func (m *RelayService) GetPayloadHeaderV1(r *http.Request, args *string, result
...
+			if err := tx.UnmarshalBinary(bytesTx); err != nil {
 				log.Println("GetPayloadHeaderV1: error decoding tx: ", err)
 				return fmt.Errorf("failed to decode tx %d: %v", i, err)
 			}
 			txs = append(txs, &tx)
+			sszTxs.Transactions = append(sszTxs.Transactions, bytesTx)
 		}
-		newRoot := types.DeriveSha(txs, trie.NewStackTrie(nil))
+
+		newRootBytes, err := sszTxs.HashTreeRoot()
```
terencechain, line 251: I recall there's a bug with this. I will check with the fssz implementers. I had handcraft my own to comply on kintsugi:

See: https://github.com/prysmaticlabs/prysm/pull/9853/files#diff-7f3786d768dbf8947d4e60b75eb8ed9f2e678f41e5a60c491c0b994fa4ab745dR103

jparyani: This is so close to perfect for me, but I'm having issues when importing `github.com/prysmaticlabs/prysm`:
```
../../.go/pkg/mod/github.com/prysmaticlabs/prysm@v1.4.2-0.20211211173600-d6338f60427f/proto/prysm/v1alpha1/generated.ssz.go:4033:7: hh.AppendBytes32 undefined (type *ssz.Hasher has no field or method AppendBytes32)
```

happens for both master and kintusgi branches of prysm. I'm going to go ahead and vendor the code for now, so thanks for pointing it out!

PR #16, opened 2021-12-18
Handle errs, nil slice declaration, and more

terencechain:
Minor cleanups, nothing substantive:
- Remove redundant casting
- Handle `RegisterService` and body close errors
- Update empty slice declarations


lib/service.go
```diff
@@ -290,7 +290,11 @@ func (m *MevService) methodNotFound(i *rpc.RequestInfo, w http.ResponseWriter) e
 		log.Println("error in method not found: doing request: ", i.Method, err)
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
```
jparyani, line 294: This seems functionally equivalent to the original version, since the error is ignored? Perhaps what we should do is log when this error happens. I’ll add it

jparyani:
Changes look great, thanks!

PR #20, opened 2021-12-22
milestone 1 spec

thegostep:


docs/milestone-1.md
```diff
@@ -0,0 +1,169 @@
+# Milestone 1 - kintsugi testnet
+
+This initial milestone provides simple middleware logic with minimal consensus client changes, simple networking, no validator authentication, and manual safety mechanism
+
+## Architecture
+
+### Block Proposal
+
+```sequence
```
terencechain, line 9: For some reason, the sequence diagram doesn't show up under the view page

docs/milestone-1.md
```diff
@@ -0,0 +1,169 @@
...
+## API Docs
+
+Methods are prefixed using the following convention:
+- `engine` prefix indicates calls made to the execution client. These methods are specified in the [execution engine APIs](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md).
+- `builder` prefix indicates calls made to the mev-boost middleware.
+- `relay` prefix indicates calls made to a relay.
+
+### engine_executePayloadV1
+
+See [engine_executePayloadV1](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md#engine_executepayloadv1).
```
terencechain, line 61: how is `engine_executePayloadV1` used in bn <-> mev_boost or mev_boost <-> relay ?

docs/milestone-1.md
```diff
@@ -0,0 +1,169 @@
...
+- randao_reveal: BLSSignature
+- eth1_data: Eth1Data
+- graffiti: Bytes32
+- proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+- attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+- attestations: List[Attestation, MAX_ATTESTATIONS]
+- deposits: List[Deposit, MAX_DEPOSITS]
+- voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+- sync_aggregate: SyncAggregate
+- execution_payload_header: ExecutionPayloadHeader
```
terencechain, line 168: Maybe it looks better if we define the type similar to spec?

```python
class BlindedBeaconBlockBody(Container):
    randao_reveal: BLSSignature
    eth1_data: Eth1Data 
    graffiti: Bytes32
    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
    attestations: List[Attestation, MAX_ATTESTATIONS]
    deposits: List[Deposit, MAX_DEPOSITS]
    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
    sync_aggregate: SyncAggregate
    execution_payload_header: ExecutionPayloadHeader
```

docs/milestone-1.md
thegostep: good catch, it is not used right now

docs/milestone-1.md
thegostep: GitHub doesn't support it unfortunately, I'll need to put the image

docs/milestone-1.md
```diff
@@ -0,0 +1,169 @@
...
+
+See [engine_forkchoiceUpdatedV1](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md#engine_forkchoiceupdatedv1).
+
+### builder_getPayloadHeaderV1
+
+#### Request
+
+- method: `builder_getPayloadHeaderV1`
+- params:
+  1. `payloadId`: `DATA`, 8 Bytes - Identifier of the payload build process
```
thegostep, line 73: Should this just be the full `PayloadAttributesV1` to avoid confusion?

docs/milestone-1.md
terencechain: Still missing the image :)

docs/milestone-1.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+## Specification
+
+1. mev-boost must be initialized with a list of (BLSPublicKey, IP address) pairs representing trusted relays.
+2. mev-boost must intercept [`engine_forkchoiceUpdatedV1`](#engine_forkchoiceupdatedv1) call from the BN -> EC and forward it to all connected relays to communicate `feeRecipient`.
+3. mev-boost must begin polling connected relays for their [`SignedMEVPayloadHeader`](#signedmevpayloadheader) using [`relay_getPayloadHeaderV1`](#relay_getpayloadheaderv1) requests.
+4. mev-boost must verify the returned [`SignedMEVPayloadHeader`](#signedmevpayloadheader) signature matches the BLS key associated with the IP of the relay and has a matching `payloadId`.
+5. upon receiving a [`builder_getPayloadHeaderV1`](#builder_getpayloadheaderv1) request from the BN, mev-boost must return the [`ExecutionPayloadHeaderV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayloadheader) with the highest associated `feeRecipientBalance`. If no eligible payload is received from a relay, mev-boost must request and return a payload from the local execution client using [`engine_getPayloadV1`](#engine_getpayloadv1).
```
terencechain, line 40: What do we think of the non-middleware solution proposed here:
https://hackmd.io/@paulhauner/H1XifIQ_t#Is-mev_boost-Middleware

Then we can remove `If no eligible payload is received from a relay, mev-boost must request and return a payload from the local execution client using [`engine_getPayloadV1`](#engine_getpayloadv1)...` and remove `engine_getPayloadV1` call below

docs/milestone-1.md
```diff
@@ -0,0 +1,175 @@
...
+- result: [`ExecutionPayloadHeaderV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayloadheader)
+- error: code and message set in case an exception happens while getting the payload.
+
+### builder_proposeBlindedBlockV1
+
+#### Request
+
+- method: `builder_proposeBlindedBlockV1`
+- params:
+  1. [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
```
tersec, line 83: This is JSON-RPC serialized, while `SignedBlindedBeaconBlock` is only specified here as an SSZ type. Aspects of SSZ type serialization aren't otherwise specified by the engine API spec on which this builds, such as `SignedBlindedBeaconBlock.signature` and `SignedMEVPayloadHeader.signature` (both `BLSSignature`s), leaving room for noninteroperability across implementations.

https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.6/src/engine/specification.md#executionpayloadv1 provides an explicit mapping between a JSON-RPC-serialization of `engine.ExecutionPayloadV1` and SSZ representation of `bellatrix.ExecutionPayload`, where both are supposed to be otherwise isomorphic.

docs/milestone-1.md
```diff
@@ -0,0 +1,175 @@
...
+#### Response
+
+- result: [`ExecutionPayloadV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayload)
+- error: code and message set in case an exception happens while proposing the payload.
+
+Technically, this call only needs to return the `transactions` field of [`ExecutionPayloadV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayload), but we return the full payload for simplicity. 
+
+### Types
+
+#### SignedMEVPayloadHeader
```
tersec, line 123: Along the lines of the comments about `SignedBlindedBeaconBlock`, but here, the JSON-RPC and SSZ types are freely intermixed, with the sort-of-SSZ-type-looking `SignedMEVPayloadHeader` containing a `MEVPayloadHeader` the fields of which are a `Data` and `Quantity` (JSON-RPC serialization types), which itself contains a `ExecutionPayloadHeaderV1`(a consensus layer SSZ type, defined using the SSZ-serialization-defined `Hash32`, `ExecutionAddress`, `Bytes32`, `uint64`, etc types). It's not internally consistent.

Implementing this might involve drawing out, at each layer, the dual JSON-RPC form of `SignedMEVPayloadHeader`, the dual SSZ form of `MEVPayloadHeader` (because the sort-of-SSZ-defined `SignedMEVPayloadHeader` either has to be read as a purely JSON-RPC object, or if not, for all its component parts to have consensus layer-like definitions in addition to the purely JSON-RPC definitions here), and likewise but inverted for the nominally-natively-SSZ `ExecutionPayloadHeaderV1`.

For example, one might imagine the CL/SSZ `MEVPayloadHeader`:
```python
class MEVPayloadHeader(Container):
    payloadHeader: ExecutionPayloadHeaderV1
    feeRecipient: ExecutionAddress
    feeRecipientBalance: uint256
```

https://github.com/realbigsean/lighthouse/blob/mev-lighthouse/beacon_node/execution_layer/src/engine_api/json_structures.rs fills this out with some gaps, but none of it is directly specified and there are various points of ambiguity.

One challenge

tersec:
Still points to CL specs 1.1.6 and engine API alpha.5. v1.1.9 and alpha.6 don't seem to impinge directly too much on this, as the `SignedBeaconBlock` aspect in particular isn't changed.

docs/milestone-1.md
```diff
@@ -0,0 +1,175 @@
...
+#### SignedBlindedBeaconBlock
+
+See [here](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/phase0/beacon-chain.md#custom-types) for the definition of fields like `BLSSignature`
+
+- `message`: [`BlindedBeaconBlock`](#blindedbeaconblock)
+- `signature`: `BLSSignature`
+
+#### BlindedBeaconBlock
+
+This is forked from [here](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/phase0/beacon-chain.md#beaconblock) with `body` replaced with `BlindedBeaconBlockBody`
```
tersec, line 147: These references to the now-Bellatrix hardfork of the CL specs don't account for, e.g., supporting multiple CL MEV-compatible CL hardforks. While I'm not sure what the role of MEV in phase 0 or Altair might be, lacking transactions and all, the next hard fork, called the sharding fork pending whatever it's renamed to, will also support transactions.

The algorithm to derive the blinded types is the same, granted -- replace the execution payload with the execution payload header in the body, then propagate that back up the dependency chain -- but this creates ambiguous naming within a single global namespace as specified. In the context of MEV boosting, is a `BlindedBeaconBlock` (signed, body, etc) the Bellatrix or Sharding version?

Of course, that won't be for a while, but it's an eventuality that will happen and it'll be smoother to anticipate it now than later.

docs/milestone-1.md
metachris: https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

docs/milestone-1.md
realbigsean: I agree this should be in JSON-RPC terms.  

We could fully specify blocks here, similar to how `ExecutionPayloadV1` is defined by the execution API, but that might not be worth doing and maintaining because we already have blocks specified [here](https://github.com/ethereum/beacon-APIs/blob/master/types/altair/block.yaml), and I don't think we'd want to leave room for discrepancy between JSON serialization of the same type between the two APIs.

I've raised a PR [here](https://github.com/ethereum/beacon-APIs/pull/194) to add Bellatrix types to the beacon API as well as endpoints and types necessary for blind transaction signing.

docs/milestone-1.md
realbigsean: AFAIK fork awareness hasn't been implemented in the engine API, and since this should be an extension of that repo I don't think it should be defined here. We should just roll with the precedent set upstream, which I imagine will be similar to what is used by the beacon API.

docs/milestone-1.md
```diff
@@ -0,0 +1,175 @@
...
+  1. [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
+
+#### Response
+
+- result: [`ExecutionPayloadV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayload)
+- error: code and message set in case an exception happens while proposing the payload.
+
+Technically, this call only needs to return the `transactions` field of [`ExecutionPayloadV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayload), but we return the full payload for simplicity.
+
+### relay_getPayloadHeaderV1
```
realbigsean, line 92: Will MEV boost forward `relay_getPayloadHeaderV1`/`relay_proposeBlindedBlockV1` requests it receives to the relay?  The format of the spec here makes it look like these are available MEV boost endpoints, but I think they might be endpoints only the relay is required to serve.  

If these are only definitions for what the relay serves, I think it would make sense to have a clearer separation (in new section at the bottom here or maybe even splitting the relay API and mev-boost API into separate docs?)

elopio:
@terencechain @realbigsean @tersec I'm sorry for the long time without a reply. We were reorganizing the team to make sure that we could maintain this for the merge.
Now @thegostep @metachris @jparyani and me will be working on this, first getting the milestone 1 tested on kiln.

Thanks a lot for your reviews and feedback. Some things are very straightforward and we are working on those. For some others I'll have more questions as I get to understand more details about the specification.

docs/milestone-1.md
elopio: we would like to move in that direction for our communication with the consensus layer. Take a look at @metachris' exploration in #53.
Also please take a look at the updated diagram in #57.

docs/milestone-1.md
elopio: Thanks @tersec and @realbigsean. I need help here to understand the proposal of using JSON-RPC for encoding these new methods. As I understand, SSZ is the new encoding for the consensus layer. I wonder why do we want to use the old encoding for the new things. Probably this has been discussed long ago, so please be patient with me :)
The way I imagine it, these APIs will be implemented by new software in the future, so these seems like a good way to start moving the future to be more consistent and easier to read. Am I being naive or missing some part of the story?

docs/milestone-1.md
elopio: oh what a mess. As I said before, I would like everything to be SSZ.

docs/milestone-1.md
tersec: SSZ is used as as the intra-consensus layer/inter-beacon-node serialization over the libp2p gossip network, yes.

JSON-RPC serializations appear in the REST API and related areas, e.g., validator clients mostly use JSON-RPC-type encodings to interact with beacon nodes. Both remain in use.

docs/milestone-1.md
elopio: Moved to https://github.com/flashbots/mev-boost/issues/67 for tracking.

docs/milestone-1.md
elopio: Moved to https://github.com/flashbots/mev-boost/issues/76 for traceability and better discussion.

elopio:
We have cleaned a milestone1 specification in https://github.com/flashbots/mev-boost/blob/main/docs/specification.md and all the pending discussions are now in the issues. Please check and let me know if I missed any concerns, to make sure we are not dropping anything important. Moving forward, we should have one issue and one PR per concern, so it's easier to focus.

Thanks for work and the nice discussion everybody. I'm closing this one.

PR #21, opened 2021-12-24
Add a simple test for payload store

terencechain:
Also changed `log.Println` + `panic` to `log.Fatalf`


jparyani:
Looks great, thanks!

PR #24, opened 2021-12-29
Use logrus

### LINKED ISSUE #23: RPC error logging
We use the standard library `log.Println` to log RPC server errors. I'd consider something more exotic like `go-ethereum/log` or `sirupsen/logrus` so we can do something like `log.Warn...`

terencechain:
Fixes #23 

This uses logrus instead of default log library to enhance loggings in `lib/service.go`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,12 +4,14 @@ go 1.16
 
 require (
 	github.com/ethereum/go-ethereum v1.10.11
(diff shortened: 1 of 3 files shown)


jparyani:
This looks great, thanks! I'm going to take this opportunity to add/fix up a few logs as well, but I'll handle it on my end

PR #34, opened 2022-01-27
Milestone 2 spec

thegostep:


docs/milestone-2.md
```diff
@@ -0,0 +1,297 @@
...
+
+- `currentState`: [`ForkchoiceStateV1`](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md#forkchoicestatev1) - current state of the chain seen by the relay.
+- `nextPayload`: [`PayloadAttributesV1`](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md#payloadattributesv1) - attributes of the upcoming payload.
+- `fraudProofs`: Array of [`FraudProof`](#fraudproof) - any fraud proof produced in the last 64 blocks.
+
+#### FraudProof
+
+- `payloadHeader`: [`SignedMEVPayloadHeader`](#signedmevpayloadheader) - the invalid payload header signed by the offending relay.
+- `payload`: [`ExecutionPayloadV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayload) - the invalid payload revealed to the network.
+- `block`: [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock) - the invalid beacon block signed and proposed by the validator.
```
terencechain, line 234: Maybe I'm missing something... What is `SignedBlindedBeaconBlock` useful for?

tbenr:
Hi!
In Teku we are start working on mev-boost integration and we have a couple of points to ask\discuss:

1. We are planning to start directly from M2 which to us seems like the authentication (the usage of `AuthenticatedFeeRecipient`) supersedes the `ProposalPromises` from M1. Do you think it is a good approach or is it too risky due to possible big changes ahead?

2. seems like you are considering the introduction on a new namespace `consensus_` that sounds like a new json-rpc endpoint. Consensus Clients do not exposes json-rpc, only REST APIs. We think that it will be wise to remain on REST and avoid to spin up a new server on a different port.

3. `AuthenticatedFeeRecipient`: since it sign only the address, it will be valid forever. I feel like it's an insecure thing to have a kind of token that is valid forever. (ie. collecting them over time and use all at once toto DOS relays via `relay_publishFeeRecipientsV1`?)

4. testing. Are some tooling available to setup a test environment with (mocked?) mev_boost to run the flows? Are there some how-to docs that can be shared?

5. Are there some dates on when an M2 testnet might be started?

docs/milestone-2.md
```diff
@@ -0,0 +1,297 @@
...
+participant consensus
+participant relays
+Title: Authentication
+mev_boost->consensus: consensus_getFeeRecipientsV1
+consensus-->mev_boost: consensus_getFeeRecipientsV1 response
+mev_boost->relays: relay_publishFeeRecipientsV1
+mev_boost->relays: relay_getValidatorStatusV1
+```
+
+1. on startup, mev-boost requests a list of [`AuthenticatedFeeRecipient`](#authenticatedfeerecipient) from the consensus client using a new [`consensus_getFeeRecipientsV1`](#consensus_getfeerecipientsv1) method.
```
realbigsean, line 20: IMO this should be an mev-boost endpoint which consensus clients must `POST` to, something like `builder_updateFeeRecipientsV1`. I don't think this API should be made bidirectional if possible. Other endpoints in the builder API should respond with a specific status code if it doesn't yet have a suggested fee recipient.  If a consensus client ever receives a payload header from mev-boost with an incorrect `suggested_fee_recipient`, they can update it with another call to `builder_updateFeeRecipientsV1`

docs/milestone-2.md
```diff
@@ -0,0 +1,297 @@
...
+    3. mev-boost begins polling connected relays for their [`SignedMEVPayloadHeader`](#signedmevpayloadheader) using [`relay_getPayloadHeaderV1`](#relay_getpayloadheaderv1) requests.
+    4. mev-boost must verify the signature and the merkle proof of the [`SignedMEVPayloadHeader`](#signedmevpayloadheader) received to ensure the signature matches the BLS key associated with the IP of the relay, has a matching `payloadId`, and has a valid fee recipient payment.
+5. upon receiving a [`builder_getPayloadHeaderV1`](#builder_getpayloadheaderv1) request from the BN, mev-boost must return the [`ExecutionPayloadHeaderV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayloadheader) with the highest associated `feeRecipientBalance` and terminate the polling cycle. If no eligible payload is received from a relay, mev-boost must request and return a payload from the local execution client using [`engine_getPayloadV1`](#engine_getpayloadv1).
```
realbigsean, line 57: > If no eligible payload is received from a relay, mev-boost must request and return a payload from the local execution client using [`engine_getPayloadV1`](#engine_getpayloadv1).
	
I have a strong preference for fallback logic like this to be solely the responsibility of the consensus client. 

Fallback logic here would necessitate mev-boost getting a full execution payload from the execution client, translating it to a blinded payload, communicating it to the consensus client, which signs and returns a blinded block, translating it back to a full signed beacon block and then publishing it via the relay... when we could just use the normal proposal flow instead (and send the block directly to p2p). I think this would be much less likely to result in a missed proposal.

Consensus clients will likely implement fallback-from-mev-boost logic regardless, so I think this endpoint should just return a status code indicating a failure as quickly as possible, and let the consensus client can handle it.

realbigsean (changes requested):

realbigsean:
>     2. seems like you are considering the introduction on a new namespace `consensus_` that sounds like a new json-rpc endpoint. Consensus Clients do not exposes json-rpc, only REST APIs. We think that it will be wise to remain on REST and avoid to spin up a new server on a different port.

Missed this when I was reviewing! but had similar thoughts here: https://github.com/flashbots/mev-boost/pull/34#discussion_r816084524

docs/milestone-2.md
```diff
@@ -0,0 +1,297 @@
...
+6. the BN must use the [`ExecutionPayloadHeaderV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#executionpayloadheader) received to assemble and sign a [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock) and return it to mev-boost using [`builder_proposeBlindedBlockV1`](#builder_proposeblindedblockv1).
+7. mev-boost must forward the [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock) to all connected relays and attach the matching [`SignedMEVPayloadHeader`](#signedmevpayloadheader) using [`relay_proposeBlindedBlockV1`](#relay_proposeblindedblockv1) to inform the network of which relay created this payload.
```
jparyani, line 59: This seems like a big hole in the security of the validation/reputation system. If we want to be able to determine when a malicious relay delays or withholds sending its `ExecutionPayloadV1`, we have to rely on mev-boost delivering the `SignedMEVPayloadHeader`s to **all** relays in a timely manner. Timing then becomes a huge edge case, since it will be impossible to tell if the relay is delayed vs the validator is delayed, or if either party lies.

metachris:
@thegostep could you remove the changes to README and milestone1 from this PR, to make it more focused? (I could also update it if you'd prefer)

docs/milestone-2.md
metachris: > I have a strong preference for fallback logic like this to be solely the responsibility of the consensus client.

This is the case in the current design, where mev-boost doesn't talk with the EL clients anymore: https://raw.githubusercontent.com/flashbots/mev-boost/main/docs/mev-boost-integration-overview.png

metachris:
As a heads-up - we're planning to close this PR and move the discussions into issues. This will go into future versions v0.3+

PR #46, opened 2022-02-28
Add the unit test coverage report 

### LINKED ISSUE #44: Add unit test coverage report
We can use codecov or coveralls to make it looks nice.

We can get the report with `go test ./lib/... ./cmd/... -v covermode=count -coverprofile=coverage.out`

elopio:
Closes #44.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,8 +19,15 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
 
-      - name: Test
-        run: make test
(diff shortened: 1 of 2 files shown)


elopio:
@metachris I think at first it didn't work because the github action was pointing to master. Also I made a mess with git and the merge to main. Then, it ran but failed because codecov didn't find the branch on their server. Then I don't know what happened, I was browsing codecov trying to avoid any log in, and then the branch appeared and the action worked :shrug: 

The thing is that now it works, I'll merge it soon, in case you want to take a look before I do it.

metachris:
Lgtm

Issue #12, opened 2021-12-16
add profit switching logic


thegostep:
see milestone 1 specification

thegostep:
Added a PR to engine API: https://github.com/ethereum/execution-apis/pull/150

thegostep:
@jparyani what is the status on this?

jparyani:
Work hasn’t begun yet, I was planning to do it alongside #9

PR #51, opened 2022-03-03
Multiple relays, profit switching and parallel request execution

metachris:
Closes:
* #7 
* #9 
* #12 

See also:
* #8 (this PR sets a hardcoded 5 second timeout)


cmd/mev-boost/main.go
```diff
@@ -21,16 +24,23 @@ var (
 	// cli flags
 	port         = flag.Int("port", defaultPort, "port for mev-boost to listen on")
 	executionURL = flag.String("executionUrl", defaultExecutionURL, "url to execution client")
-	relayURL     = flag.String("relayUrl", defaultRelayURL, "url to relay")
+	relayURL     = flag.String("relayUrl", defaultRelayURL, "url to relay (single entry or comma-separated list)")
```
jparyani, line 27: rename everywhere to `relayUrls`, including the CLI param

metachris, line 27: change to `relayUrls`? 👍  👎

metachris: will do. wasn't sure because that's a breaking change

cmd/mev-boost/main.go
```diff
@@ -16,21 +19,28 @@ var (
 	// defaults
 	defaultPort         = 18550
 	defaultExecutionURL = getEnv("EXECUTION_URL", "http://127.0.0.1:18545")
-	defaultRelayURL     = getEnv("RELAY_URL", "http://127.0.0.1:28545")
+	defaultRelayURL     = getEnv("RELAY_URLS", "http://127.0.0.1:28545")
```
bhakiyakalimuthu, line 22: To make it consistent ?
```suggestion
	defaultRelayURLs     = getEnv("RELAY_URLS", "http://127.0.0.1:28545")
```

lib/service.go
```diff
@@ -72,52 +65,86 @@ func makeRequest(url string, method string, params []interface{}) ([]byte, error
...
+		}(url)
+	}
+
+	// Collect responses from the goroutines
+	for i := 0; i < cap(results); i++ {
+		res := <-results
+		ret[res.url] = res
+	}
+
+	return ret
```
bhakiyakalimuthu, line 102: I think you don't need to return `ret`..bcz ret is already a  named return value. I have seen this in many of our repos(Just noticed but not a problem)
```suggestion
	return
```

lib/service.go
```diff
@@ -72,52 +65,86 @@ func makeRequest(url string, method string, params []interface{}) ([]byte, error
...
 }
 
 // ForkchoiceUpdatedV1 TODO
 func (m *MevService) ForkchoiceUpdatedV1(r *http.Request, args *[]interface{}, result *ForkChoiceResponse) error {
```
bhakiyakalimuthu, line 106: I would consider changing this to make it consistent with `ForkChoiceResponse`
```suggestion
func (m *MevService) ForkChoiceUpdatedV1(r *http.Request, args *[]interface{}, result *ForkChoiceResponse) error {
```

metachris: This is okay as is - in the spec the function is called `engine_forkchoiceUpdatedV1` (https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#engine_forkchoiceupdatedv1)

lib/service.go
```diff
@@ -130,126 +169,173 @@ func (m *RelayService) ProposeBlindedBlockV1(r *http.Request, args *SignedBlinde
...
+			logMethod.WithFields(logrus.Fields{"error": err, "data": string(res.res.Result)}).Error("Could not unmarshal response")
+			continue
+		}
+
+		logMethod.WithFields(logrus.Fields{
+			"blockHash": result.BlockHash,
+			"number":    result.Number,
+			"txRoot":    fmt.Sprintf("%#x", result.TransactionsRoot),
+		}).Info("ProposeBlindedBlockV1: revealed new payload from relay")
+		return nil
```
Ruteri, line 216: Nit: you could cancel the rest of the requests too. I guess now is not a good time, but it'd be nice to have it at some point.

lib/service.go
```diff
@@ -130,126 +169,173 @@ func (m *RelayService) ProposeBlindedBlockV1(r *http.Request, args *SignedBlinde
...
-				logMethod.WithField("err", err).Error("Mismatched tx root")
+		// Skip processing this result if lower fee than previous
+		if result != nil && result.FeeRecipientDiff != nil {
+			if _result.FeeRecipientDiff == nil || _result.FeeRecipientDiff.Cmp(result.FeeRecipientDiff) < 1 {
+				continue
+			}
+		}
+
+		// Use this relay's response as mev-boost response because it's most profitable
+		*result = *_result
```
Ruteri, line 280: Can `result` be nil here? If not, then I think the check in previous if is a bit confusing.

lib/store.go
```diff
@@ -40,13 +47,27 @@ func (s *store) Get(blockHash common.Hash) *ExecutionPayloadWithTxRootV1 {
...
+	s.forkchoiceMutex.RLock()
+	defer s.forkchoiceMutex.RUnlock()
+	forkchoiceResponses, found := s.forkchoices[payloadID]
+	return forkchoiceResponses, found
+}
+
+func (s *store) SetForkchoiceResponse(boostPayloadID, relayURL, relayPayloadID string) {
+	s.forkchoiceMutex.Lock()
+	defer s.forkchoiceMutex.Unlock()
+	s.forkchoices[boostPayloadID] = make(map[string]string)
```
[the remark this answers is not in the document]
metachris: Yes, single boostPayloadID, multiple relay URLs + their respective payloadIDs

cmd/mev-boost/main.go
```diff
@@ -14,21 +17,28 @@ var (
...
 
 	// cli flags
-	port     = flag.Int("port", defaultPort, "port for mev-boost to listen on")
-	relayURL = flag.String("relayUrl", defaultRelayURL, "url to relay")
+	port      = flag.Int("port", defaultPort, "port for mev-boost to listen on")
+	relayURLs = flag.String("relayUrl", defaultRelayURLs, "relay urls - single entry or comma-separated list")
 )
 
 func main() {
+	rand.Seed(time.Now().UnixNano())
```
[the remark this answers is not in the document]
metachris: Why? Or what would be the alternative? We just need this to create random payloadIDs.

lib/service.go
```diff
@@ -61,41 +58,83 @@ func makeRequest(url string, method string, params []interface{}) ([]byte, error
...
+			logMethod.WithFields(logrus.Fields{"error": err, "data": string(res.res.Result)}).Error("Could not unmarshal response")
+			continue
+		}
+
+		if forkchoiceResponse.Status != ForkchoiceStatusValid && forkchoiceResponse.Status != "SUCCESS" && forkchoiceResponse.Status != "" { // SUCCESS is used by mergemock, although it's not in the engine spec (also accept empty status because mergemock)
+			logMethod.WithFields(logrus.Fields{"error": res.err, "url": res.url, "status": forkchoiceResponse.Status}).Warn("status not valid")
+			continue
+		}
+
+		m.store.SetForkchoiceResponse(boostPayloadID.String(), res.url, forkchoiceResponse.PayloadID.String())
```
terencechain, line 125: I'd probably check `forkchoiceResponse.PayloadID != nil` just to be safe. In the execution engine spec, you could have the following case:
```
{payloadStatus: {status: VALID, latestValidHash: forkchoiceState.headBlockHash, validationError: null}, payloadId: null} if the payload is deemed VALID and a build process hasn't been started
```

cmd/mev-boost/main.go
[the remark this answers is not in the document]
Ruteri: Sure, for rpc IDs this is good enough, just not really a secure seed.
I'd comment out that this shouldn't be used where cryptographically secure random numbers are needed.

lib/store.go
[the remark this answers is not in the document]
metachris: fixed

lib/service.go
metachris: Nope, result is never nil. And yeah, good point, removed nil check in previous if.

lib/service.go
metachris: Great input, thanks! Updated

lib/service.go
metachris: Cool idea! Added a todo and we could revisit this later :)

lib/service.go
```diff
@@ -61,41 +58,83 @@ func makeRequest(url string, method string, params []interface{}) ([]byte, error
...
+		}(url)
 	}
 
-	err = json.Unmarshal(resp.Result, result)
-	if err != nil {
-		logMethod.WithField("err", err).Error("Could not unmarshal response", method, err)
-		return err
+	// Collect the responses
+	hasValidResponse := false
+	for i := 0; i < cap(resultC); i++ {
```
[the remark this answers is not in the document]
metachris: done

jparyani (approved):

PR #55, opened 2022-03-13
Remove executePayloadV1, add getPayloadV1

terencechain:
Remove deprecated `engine_executePayloadV1` and add the missing `engine_getPayloadV1` call. I also updated execution-apis version to kilnv2 to better align with #47


jparyani:
Looks good, thanks!

PR #56, opened 2022-03-15
readme: mermaid sequence diagram

metachris:
Removes EL client, otherwise same flow as `docs/block-proposal.png`. Should probably replace `docs/block-proposal.png` at some point.


jparyani:
Very cool! I wasn’t aware of mermaid support on github.

PR #57, opened 2022-03-15
update docs/block-proposal.png based on mermaid

metachris:


elopio (approved):
Thank you @metachris, looks beautiful.

PR #58, opened 2022-03-18
Add `ForkchoiceResponse` cache test

terencechain:
- Fixed a few typos
- Return error for `rand.Read(boostPayloadID)`
- Add a test for the `ForkchoiceResponse` cache


metachris (approved):
Very cool, thank you 👍

PR #59, opened 2022-03-22
Rm `engine_getPayloadV1` and update versions

terencechain:
I don't think `engine_getPayloadV1` is relevant in the latest proposal. I also updated the version to the latest `v1.1.10` for consensus spec and `v1.0.0-alpha.8` for API spec


elopio:
Thank you @terencechain!

PR #66, opened 2022-03-30
store: periodic cleanup

### LINKED ISSUE #54: store: clean up periodically
Currently, the [store](https://github.com/flashbots/mev-boost/blob/main/lib/store.go) is accumulating but not freeing.

metachris:
Not sure where to kickoff the loop that executes the periodic cleanup. Right now I've added it to `NewStore`:

```go
func NewStore(startCleanupLoop bool) Store {
```

Would love to hear thoughts/feedback on what's the best place and way to do this.

Store entries are kept around for a minimum of 7 minutes, just a bit more than an epoch with 6.4 min.

Closes #54

### DIFF
--- a/cmd/mev-boost/main.go
+++ b/cmd/mev-boost/main.go
@@ -37,7 +37,7 @@ func main() {
 		_relayURLs = append(_relayURLs, strings.Trim(entry, " "))
 	}
 
-	store := lib.NewStore()
+	store := lib.NewStore(true)
 	router, err := lib.NewRouter(_relayURLs, store, log)
 	if err != nil {
 		panic(err)
--- a/cmd/mev-boost/main.go
+++ b/cmd/mev-boost/main.go
@@ -37,7 +37,7 @@ func main() {
 		_relayURLs = append(_relayURLs, strings.Trim(entry, " "))
 	}
 
-	store := lib.NewStore(true)
+	store := lib.NewStoreWithCleanup()
 	router, err := lib.NewRouter(_relayURLs, store, log)
 	if err != nil {
 		panic(err)
--- a/lib/store.go
+++ b/lib/store.go
@@ -9,7 +9,13 @@ import (
 
 var (
 	cleanupLoopInterval = 5 * time.Minute
-	stateExpiry         = 7 * time.Minute // a bit more than an epoch with 6.4 min
+
+	secondsPerSlot = 12
+	slotsPerEpoch  = 32
+	stateExpiry    = time.Second * time.Duration(secondsPerSlot*slotsPerEpoch*2) // ~2 epochs
+
+	// local now function, used instead of time.Now so it can be overwritten in tests
+	now = time.Now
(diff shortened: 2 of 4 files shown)


jparyani (approved):

cmd/mev-boost/main.go
```diff
@@ -37,7 +37,7 @@ func main() {
 		_relayURLs = append(_relayURLs, strings.Trim(entry, " "))
 	}
 
-	store := lib.NewStore()
+	store := lib.NewStore(true)
```
terencechain, line 40:
```suggestion
	store := lib.NewStore(true /* start clean up loop /*)
```

lib/store.go
```diff
@@ -2,37 +2,73 @@ package lib
...
 import (
 	"sync"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
+var (
+	cleanupLoopInterval = 5 * time.Minute
+	stateExpiry         = 7 * time.Minute // a bit more than an epoch with 6.4 min
```
terencechain, line 12: Have we thought of using hard-coded configs rather than a magic number?  If any seconds per slot or slots per epoch change, this will be hard to find

```go
stateExpiry = slotsPerEpoch * secondsPerSlot * time.Second + someBuffer
```

lib/store.go
```diff
@@ -55,21 +91,42 @@ func (s *store) SetExecutionPayload(blockHash common.Hash, payload *ExecutionPay
...
 	s.forkchoiceMutex.Lock()
 	defer s.forkchoiceMutex.Unlock()
 	if _, ok := s.forkchoices[boostPayloadID]; !ok {
-		s.forkchoices[boostPayloadID] = make(map[string]string)
+		s.forkchoices[boostPayloadID] = newForkchoiceResponseContainer()
+	}
+	s.forkchoices[boostPayloadID].Payload[relayURL] = relayPayloadID
+}
+
+// Cleanup removes all payloads older than 7 minutes (a bit more than an epoch, which is 6.4 minutes)
```
terencechain, line 113: https://github.com/patrickmn/go-cache is another good TTL cache, it's probably overkill for now, maybe for the future : )

cmd/mev-boost/main.go
terencechain: Rather than passing in boolean flag, another approach is:

```go
func NewStore() Store {
  return s
}

func NewStoreWithCleanup() Store {
  s := NewStore()
  // Start clean up routine
  return s
}
```

terencechain (commented):
Nice!

cmd/mev-boost/main.go
metachris: I like it!

metachris: updated

lib/store.go
metachris: updated

Issue #71, opened 2022-03-30
Remove use of `engine_forkchoiceUpdatedV1`


lightclient:
Currently `engine_forkchoiceUpdatedV1` is used to communicate to relays the `feeRecipient` that a validator will be using. As a side effect, a lot of unnecessary information is communicated to the relays.

I propose two potential alternatives:

1) `builder_getPayloadHeader(hash, feeRecipient) -> PayloadHeader`
2) `builder_preparePayload(hash, feeRecipient) -> uint64` + `builder_getPayloadHeader(payloadId) -> PayloadHeader`

I don't know if I have enough of an understanding of the latency of 1) to choose it over 2), but it is naively my preference.

mkalinin:
I think that entire `PayloadAttributes` object is worth being communicated to relays. Ofc, this information is available in the network, but it probably would be easier for builders to receive it from the outside as in this case builders will not have to query CL to obtain this information. Also, querying CL isn't that easy to do in general case, suppose, there are two competing forks, a builder will have to understand which fork a validator is proposing on and make the corresponding query to CL -- and all information the builder has is `hash`, so, it will query CL by execution block hash which may be inconvenient.

I also think that it might worth communicating the attributes in advance as per (2). In this case `feeRecipient` and other payload attributes that may affect the result of transaction execution will be known to builders in advance aiding creation of a better block.

elopio:
From @djrtwo in https://notes.ethereum.org/WX1hwcMRRzy5oh-zHyzqOA?view

# use of forkchoiceUpdated

    reuse of the engine_forkchoiceUpdatedV1 semantics seems off
    I understand that some of the same info is to be communicated (e.g. payloadBuildAttributes) but to use the same namespace engine_ leads to a less clear separation of concerns and could lead to misunderstandings or accidental bugs
    the description of the namespaces (under header API) conflicts with actual usage (i.e. engine is one of the methods called to the sidecar). suggest again for there to be a new builder_ endpoint that initiates the mev-build process rather than reusing engine_forkchoiceUpdated
    I would suggest builder_beginPayloadBuildProcessV1 for BN->mev_boost and relay_begin_payloadBuildProcessV1 for mev_boost->relay
```
builder_beginPayloadBuildProcessV1 {
    headBlockHash,
    payloadAttributes
}

relay_beginPayloadBuildProcessV1 {
    headBlockHash,
    payloadAttributes
}

# if relay cannot build on requested headBlockHash, then can return null
[...]

PR #79, opened 2022-04-02
remove use of engine_forkchoiceUpdated, send hash+feeRecipient with getPayloadHeader

metachris:
Code proposal for option 1 in #71 (remove use of `engine_forkchoiceUpdated`)
Updated `getPayloadHeaderV1` method params to `[headBlockHash, feeRecipient]`

Sequence without fcU:

```mermaid
sequenceDiagram
    Title: Block Proposal
    consensus->>+mev_boost: builder_getPayloadHeaderV1
    mev_boost->>relays: relay_getPayloadHeaderV1
    Note over mev_boost: select most valuable payload
    mev_boost-->>-consensus: builder_getPayloadHeaderV1 response
    Note over consensus: sign the block
    consensus->>+mev_boost: builder_proposeBlindedBlockV1
    Note over mev_boost: identify payload source
    mev_boost->>relays: relay_proposeBlindedBlockV1
    Note over relays: validate signature
    relays-->>mev_boost: relay_proposeBlindedBlockV1 response
    mev_boost-->>-consensus: builder_proposeBlindedBlockV1 response
```

If this is the way we decide to go, I'm happy to update the spec accordingly as well.


metachris:
Fyi, probably not going to go this route, because the delay might be an issue (`builder_getPayloadHeader` needs to give the relays sufficient time to build the payload), and we might want/need to cache the payload already at mev-boost 🤔

PR #82, opened 2022-04-04
Simplify spec, update based on recent discussions

lightclient:
I took a stab at updating the spec based on our recent discussions about the spec. I need to update sequence diagram and the process overview, but I think it's relatively straightforward.

A few notes: 
* I removed the concept of `builder` vs. `relay` methods and combined them. The CL can ignore things like `feeRecipientDiff` and signatures, or it can use that info to compare against it's local payload. I think this really simplifies things a lot.
* I decided to go with SSZ encoding the block because it simplifies the spec a lot and is already the form that the builder will need it in to validate the signature.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -2,13 +2,39 @@
 
 This document specifies the Builder API methods that the Consensus Layer uses to interact with external block builders.
 
+```mermaid
+sequenceDiagram
+    participant consensus
+    participant mev_boost
+    participant relays
+    Title: Block Proposal
+    Note over consensus: sign fee recipient announcement
+    consensus->>mev_boost: builder_setFeeRecipient
+    mev_boost->>relays: builder_setFeeRecipient
+    Note over consensus: wait for allocated slot
+    consensus->>mev_boost: builder_getHeader
+    mev_boost->>relays: builder_getHeader
+    relays-->>mev_boost: builder_getHeader response
+    Note over mev_boost: verify response matches expected
+    Note over mev_boost: select best payload
+    mev_boost-->>consensus: builder_getHeader response
+    Note over consensus: sign the block
+    consensus->>mev_boost: builder_getPayload
+    Note over mev_boost: identify payload source
+    mev_boost->>relays: builder_getPayload
+    Note over relays: validate signature
+    relays-->>mev_boost: builder_getPayload response
+    Note over mev_boost: verify response matches expected
+    mev_boost-->>consensus: builder_getPayload response
+```
+
 ## Structures
 
 ### `ExecutionPayloadV1`
 
 Mirror of [`ExecutionPayloadV1`][execution-payload].
 
-### `BlindExecutionPayloadV1`
+### `ExecutionPayloadHeaderV1`
 
 Equivalent to `ExecutionPayloadV1`, except `transactions` is replaced with `transactionsRoot`.
 - `parentHash`: `DATA`, 32 Bytes
@@ -26,6 +52,55 @@ Equivalent to `ExecutionPayloadV1`, except `transactions` is replaced with `tran
 - `blockHash`: `DATA`, 32 Bytes
 - `transactionsRoot`: `DATA`, 32 Bytes
 
+#### SSZ Objects
+
+Consider the following definitions supplementary to the definitions in [`consensus-specs`][consensus-specs].
+
+##### `builder_setFeeRecipientV1` Request
+
+```python
+class Request(Container):
+    feeRecipient: Bytes20
+    timestamp: uint64
+```
+
+##### `builder_getPayloadV1` Response
+
+```python
+class Response(Container):
+    payload: ExecutionPayloadHeader
+    value: uint256
+```
+
+##### `builder_getPayloadV1` Request
+
+###### `BlindBeaconBlock`
+
+```python
+class BlindBeaconBlock(Container):
+    slot: Slot
+    proposer_index: ValidatorIndex
+    parent_root: Root
+    state_root: Root
+    body: BlindBeaconBlockBody
+```
+
+###### `BlindBeaconBlockBody`
+
+```python
+class BlindBeaconBlockBody(Container):
+    randao_reveal: BLSSignature
+    eth1_data: Eth1Data
+    graffiti: Bytes32
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    sync_aggregate: SyncAggregate
+    execution_payload: ExecutionPayloadHeader
+```
+
 ## Errors
 
 The list of error codes introduced by this specification can be found below.
@@ -36,7 +111,7 @@ The list of error codes introduced by this specification can be found below.
 | -32001 | Unknown hash | No block with the provided hash is known. |
 | -32002 | Unknown validator | No known mapping between validator and feeRecipient. |
 | -32003 | Invalid SSZ | Unable to decode SSZ. |
-| -32004 | Unknown block | Block does not match the provided blind header. |
+| -32004 | Unknown block | Block does not match the provided header. |
 | -32005 | Invalid signature | Provided signature is invalid. |
 | -32006 | Invalid timestamp | Provided timestamp was invalid. |
 | -32600 | Invalid request | The JSON sent is not a valid Request object. |
@@ -52,24 +127,10 @@ The list of error codes introduced by this specification can be found below.
 All signature operations should follow the [standard BLS operations][bls] interface defined in `consensus-specs`.
 
 There are two types of data to sign over in the Builder API:
-* In-protocol messages, e.g. [`BeaconBlock`][beacon-block], which should compute the signing root using [`computer_signing_root`][compute-signing-root] and use the domain specified for beacon block proposals.
+* In-protocol messages, e.g. [`BlindBeaconBlock`](#blindbeaconblock), which should compute the signing root using [`computer_signing_root`][compute-signing-root] and use the domain specified for beacon block proposals.
 * Builder API messages, e.g. [`builder_setFeeRecipientV1`](#builder_setFeeRecipientV1), which should compute the signing root using [`compute_signing_root`][compute-signing-root] and the domain `DomainType('0xXXXXXXXX')` (TODO: get a proper domain).
 
-As `compute_signing_root` takes `SSZObject` as input, client software should convert in-protocol messages to their SSZ representation to compute the signing root and Builder API messages to the SSZ representations defined below. See [`consensus-specs`][consensus-specs] for the definitions of [`Bytes20`][bytes20], [`ExecutionPayloadHeader`][execution-payload-header], and [`uint256`][uint256].
-
-#### `feeRecipient`
-
-```python
-feeRecipient = Bytes20
-```
-
-#### `builder_getExecutionPayloadV1` Response
-
-```python
-class Response(Container):
-    payload: ExecutionPayloadHeader
-    value: uint256
-```
+As `compute_signing_root` takes `SSZObject` as input, client software should convert in-protocol messages to their SSZ representation to compute the signing root and Builder API messages to the SSZ representations defined [above](#sszobjects).
 
 ## Methods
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -1,153 +1,141 @@
...
+### `builder_setFeeRecipientV1`
 
 #### Request
 
-- method: `builder_getPayloadHeaderV1`
+- method: `builder_setFeeRecipientV1`
 - params:
-  1. `payloadId`: `DATA`, 8 Bytes - Identifier of the payload build process
+  1. `feeRecipient`: `DATA`, 20 Bytes - Address of account which should receive fees.
+  2. `signature`: `DATA`, 96 Bytes - Validator signature over `feeRecipient`.
```
jparyani, line 82: Is this signature only on the `feeRecipient`? This seems very unsafe since it's easy to leak and masquerade as another validator. Adding either some id for the relay and/or a timing component would fix this I think.

mkalinin: There is simply no way to recover a public key in BLS schema. It should be a part of signed data.

metachris: A timestamp would be good! So the relays can build for the latest sent one.

metachris:
Thanks for putting this together @lightclient! Good stuff in there. I'd love to have `setFeeRecipient` in the launch version, and +1 on SSZ.

I think there's going to be a bit more discussion about removing builder/relay separation, because having a relay prefix allows for a bit more flexibility on iteration.

Will follow up tomorrow.

docs/specification.md
lightclient: @jparyani how would this allow someone to masquerade as another validator? The builder would map `feeRecipient` to the validator's public key. Since the pk is verified by the signature, there is no malleability. People on other IPs could resubmit the message, but there shouldn't be an issue with that.

@mkalinin thanks for noting this, keep forgetting to add that in.

@metachris good suggestion, I will update to include this.

docs/specification.md
```diff
@@ -1,153 +1,146 @@
...
+### `builder_getExecutionPayloadV1`
 
 #### Request
 
-- method: `relay_proposeBlindedBlockV1`
+- method: `builder_getExecutionPayloadV1`
 - params:
-  1. [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
-  2. [`SignedMEVPayloadHeader`](#signedmevpayloadheader)
+  1. `block`: `DATA`, arbitray length - SSZ encoded [`BeaconBlock`][beacon-block].
```
realbigsean, line 125: This has to be a `BlindedBeaconBlock` (an otherwise full beacon block with a blinded execution payload) because the CL has no knowledge of transactions at this point, just the transactions root

lightclient: Absolutely, my mistake - was just fixing that.

elopio:
> I think there's going to be a bit more discussion about removing builder/relay separation, because having a relay prefix allows for a bit more flexibility on iteration.

I think the relay concept is outdated, we can get rid of it. mev-boost now requests full blocks to a set of builders, and chooses the most profitable.
Or mev-boost could talk to only one builder aggregator that sends a single proposal. This aggregator can conform to the same API of a single builder.

We can move forward this way, and if the relay concept becomes relevant we reintroduce it. I agree with @lightclient that this clarifies concepts and simplifies the specification. @metachris I wonder what kind of iteration do you have in mind here.

metachris:
I like this proposal and am on board with it. 

Side-question: I wonder if mev-boost should just poll the relays always for headers, so it can return a payload very quickly when the slot has arrived (the relay could deliver a header for every fork choice even a few blocks deep).

realbigsean:
In order for mev-boost to have a role in the reputation management of relays/builders (in my view they're essentially one in the same), won't we still need an equivalent to the relay namespace? Because there will be information required by mev-boost related to management that the CL should never need from mev-boost

docs/specification.md
```diff
@@ -1,153 +1,207 @@
...
 
-### builder_proposeBlindedBlockV1
+##### `builder_setFeeRecipientV1` Request
 
-#### Request
+```python
+class Request(Container):
+    feeRecipient: Bytes20
+    timestamp: uint64
+```
```
terencechain, line 65: I think we are missing public key and signature here. As seen below:
```
  1. `feeRecipient`: `DATA`, 20 Bytes - Address of account which should receive fees.
  2. `timestamp`: `QUANTITY`, uint64 - Unix timestamp of announcement.
  3. `publicKey`: `DATA`, 48 Bytes - Public key of validator.
  4. `signature`: `DATA`, 96 Bytes - Signature over `feeRecipient`.
  ```
  
Or do we neglect the rest because it's not part of signing root? in that case, `timestamp` should be removed?

File: docs/specification.md [hunk 4]
```diff
@@ -1,153 +1,207 @@
...
 
-- method: `relay_proposeBlindedBlockV1`
+- method: `builder_setFeeRecipientV1`
 - params:
-  1. [`SignedBlindedBeaconBlock`](#signedblindedbeaconblock)
-  2. [`SignedMEVPayloadHeader`](#signedmevpayloadheader)
+  1. `feeRecipient`: `DATA`, 20 Bytes - Address of account which should receive fees.
+  2. `timestamp`: `QUANTITY`, uint64 - Unix timestamp of announcement.
+  2. `publicKey`: `DATA`, 48 Bytes - Public key of validator.
+  3. `signature`: `DATA`, 96 Bytes - Signature over `feeRecipient`.
```
terencechain, line 146:
```suggestion
  3. `publicKey`: `DATA`, 48 Bytes - Public key of validator.
  4. `signature`: `DATA`, 96 Bytes - Signature over `feeRecipient`.
```

docs/specification.md
```diff
@@ -1,153 +1,207 @@
...
+### `builder_getHeaderV1`
 
-#### MEVPayloadHeader
+#### Request
 
-- payloadHeader: [`ExecutionPayloadHeaderV1`](https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/beacon-chain.md#executionpayloadheader)
-- feeRecipientDiff: Quantity, 256 Bits - the change in balance of the feeRecipient address
+- method: `builder_getHeaderV1`
+- params:
+  1. `hash`: `DATA`, 32 Bytes - Hash of block which the validator intends to use as the parent for its proposal.
```
terencechain, line 164: should the validator public key be part of the request params or else how does the builder return `Unknown validator` if the validator public key has not been mapped to a `feeRecipient`?

docs/specification.md
```diff
@@ -1,153 +1,207 @@
...
 
-#### SignedBlindedBeaconBlock
+#### Response
 
-See [here](https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/phase0/beacon-chain.md#custom-types) for the definition of fields like `BLSSignature`
+- result: `object`
+    - `payload`: [`ExecutionPayloadHeaderV1`](#executionpayloadheaderv1).
```
terencechain, line 169: Better?

```suggestion
    - `header`: [`ExecutionPayloadHeaderV1`](#executionpayloadheaderv1).
```

docs/specification.md
```diff
@@ -1,153 +1,207 @@
...
 
-- method: `relay_getPayloadHeaderV1`
-- params:
-  1. `payloadId`: `DATA`, 8 Bytes - Identifier of the payload build process
+All signature operations should follow the [standard BLS operations][bls] interface defined in `consensus-specs`.
 
-#### Response
+There are two types of data to sign over in the Builder API:
+* In-protocol messages, e.g. [`BlindBeaconBlock`](#blindbeaconblock), which should compute the signing root using [`computer_signing_root`][compute-signing-root] and use the domain specified for beacon block proposals.
+* Builder API messages, e.g. [`builder_setFeeRecipientV1`](#builder_setFeeRecipientV1), which should compute the signing root using [`compute_signing_root`][compute-signing-root] and the domain `DomainType('0xXXXXXXXX')` (TODO: get a proper domain).
```
terencechain, line 131: `builder_getHeaderV1`'s response requires signing too

docs/specification.md
```diff
@@ -1,153 +1,207 @@
...
 
-- slot: Slot
-- proposer_index: ValidatorIndex
-- parent_root: Root
-- state_root: Root
-- body: BlindedBeaconBlockBody
+- method: `builder_getPayloadV1`
+- params:
+  1. `block`: `DATA`, arbitray length - SSZ encoded [`BlindBeaconBlock`](#blindbeaconblock).
+  2. `signature`: `DATA`, 96 Bytes - BLS signature of the validator over `block`.
```
terencechain, line 187: Another option is to pass in `SignedBlindBeaconBlock` which is basically 1. + 2.

Have we considered adding `publicKey` as an input? We don't need it if the builder has the beacon state and the ability to look up the `block. proposer_index`'s public key

docs/specification.md
metachris: > 4. `signature`: `DATA`, 96 Bytes - Signature over `feeRecipient`.

Should this be "Signature over `feeRecipient + timestamp`?

docs/specification.md
lightclient: I'm not sure the best way to go about specify all the different types here. The SSZ types are generally only used to calculate the signing root, so that's why they don't include the public keys or signatures. There is a mistake in the `builder_setFeeRecipientV1` method where it should say the signature is over `feeRecipient` and `timestamp`. I will fix this. I welcome any suggestions on how to make the specification clearer.

> In that case, timestamp should be removed?

Could you clarify why the timestamp should be removed? If `timestamp` is malleable, it could be difficult for a validator to change their fee recipient, because attackers could replay old signatures.

docs/specification.md
lightclient: Yes, thank you for pointing out. I will update.

docs/specification.md
lightclient: Update :+1:

docs/specification.md
lightclient: My expectation was that the builder would know what validator is planning to propose for the current slot and would return `Unknown validator` if it had not received a `setFeeRecipient` message from the validator. I will update the wording here, because "recovered" doesn't make much sense in this context.

This does make me wonder if the builder should also return the slot the slot it is building for in case there is a mismatch?

docs/specification.md
lightclient: As mentioned in this [comment](https://github.com/flashbots/mev-boost/pull/82/files#r845003459), I am making the assumption that builder's have the ability to look up the `block.proposer_index`'s public key. I think this is a reasonable and important assumption, because without it, the builder may receive may `builder_getHeader` requests with different `feeRecipient`s, which will cause them to calculate blocks that have no possibility of being included.

I'm not necessarily opposed to passing `SignedBlindBeaconBlock` - the rationale for not doing so currently is it breaks from the current pattern of `signature` always being an RPC parameter / response.

lightclient:
> In order for mev-boost to have a role in the reputation management of relays/builders (in my view they're essentially one in the same), won't we still need an equivalent to the relay namespace? Because there will be information required by mev-boost related to reputation management that the CL should never need from mev-boost

@realbigsean: could you give an example? So far, I'm trying to specify each method in way that makes faults attributable (the thorny unattributable behavior being the builder withholding data). From there, my understanding is that `mev-boost` clients will need to gossip amongst themselves this information, so it wouldn't be a direct communication between CL<>`mev-boost` or builder<>`mev-boost`.

docs/specification.md
terencechain: > Could you clarify why the timestamp should be removed? If timestamp is malleable, it could be difficult for a validator to change their fee recipient, because attackers could replay old signatures.

Oh no, I don't think it should. I think it's because we missed mentioning `timeStamp` under `signature: DATA, 96 Bytes - Signature over feeRecipient.`, that's why I thought it was not required here. We are good here!

docs/specification.md
terencechain: There could skip slots on top of `hash`
```
hash -> request/propose
hash -> skipped_slot -> request/propoase
hash -> skipped_slot -> skipped_slot -> request/propose
```
The builder may utilize the current wall block but if there's a tiny clock synchronization mismatch between the BN and builder then it may fall apart. It may be safer to include an additional `slot` or `proposer_index` or `pub_key` in the request as an additional anchor point

docs/specification.md
terencechain: No worries. I think `block` and `signature` are great. Let's see if others have feedback on this

realbigsean:
> From there, my understanding is that mev-boost clients will need to gossip this information amongst themselves, so it wouldn't be a direct communication between CL<>mev-boost or builder<>mev-boost and therefore does not necessitate a method.

Yes I'm thinking about how exactly instances of `mev-boost` can communicate issues to each other. I think a medium term goal would be to include something like fraud proofs into consensus layer p2p ... but as far as how this works at the merge, are there plans for `mev-boost` to implement some type of p2p communication?  

I was under the impressing there would be something similar to what was described here https://github.com/flashbots/mev-boost/pull/34
Where relays/builders are relaying information about fraud proofs as well. It would work by `mev-boost` broadcasting fraud proofs to all relays/builders, and relays/builder making fraud proofs available to any other connected `mev-boost` clients. This would be reliant on relays/builders policing each other, I doubt an individual relay/builder would make available fraud proofs for payloads it produced

metachris:
Merged this as new spec v0.2! Thanks for all the contributions ⭐ 

Let's move any remaining and further discussions to separate issues and PRs.

docs/specification.md
metachris: @terencechain thanks for the brainstorming on this. the PR is now merged, please open an issue/PR to continue the thread, if applicable 🙏

sambacha:
> > From there, my understanding is that mev-boost clients will need to gossip this information amongst themselves, so it wouldn't be a direct communication between CL<>mev-boost or builder<>mev-boost and therefore does not necessitate a method.
> 
> Yes I'm thinking about how exactly instances of `mev-boost` can communicate issues to each other. I think a medium term goal would be to include something like fraud proofs into consensus layer p2p ... but as far as how this works at the merge, are there plans for `mev-boost` to implement some type of p2p communication?
> 
> I was under the impression there would be something similar to what was described here #34 Where relays/builders are relaying information about fraud proofs as well. It would work by `mev-boost` broadcasting fraud proofs to all relays/builders, and relays/builder making fraud proofs available to any other connected `mev-boost` clients. This would be reliant on relays/builders policing each other, I doubt an individual relay/builder would make available fraud proofs for payloads it produced

Couldn't the attnets attribute in the node's ENR be used for initially determining if there is an issue with the node? Since the value of the attnets attribute is a bitvector showing what subnet the node is subscribed to, faulty nodes would show up, etc etc?

Issue #76, opened 2022-03-31
SignedMEVPayloadHeader mixes JSON-RPC and SSZ types


elopio:
[tersec](https://github.com/tersec) [on Feb 2](https://github.com/flashbots/mev-boost/pull/20#discussion_r797539182) •

Along the lines of the comments about SignedBlindedBeaconBlock (#69), but here, the JSON-RPC and SSZ types are freely intermixed, with the sort-of-SSZ-type-looking SignedMEVPayloadHeader containing a MEVPayloadHeader the fields of which are a Data and Quantity (JSON-RPC serialization types), which itself contains a ExecutionPayloadHeaderV1(a consensus layer SSZ type, defined using the SSZ-serialization-defined Hash32, ExecutionAddress, Bytes32, uint64, etc types). It's not internally consistent.

Implementing this might involve drawing out, at each layer, the dual JSON-RPC form of SignedMEVPayloadHeader, the dual SSZ form of MEVPayloadHeader (because the sort-of-SSZ-defined SignedMEVPayloadHeader either has to be read as a purely JSON-RPC object, or if not, for all its component parts to have consensus layer-like definitions in addition to the purely JSON-RPC definitions here), and likewise but inverted for the nominally-natively-SSZ ExecutionPayloadHeaderV1.

For example, one might imagine the CL/SSZ MEVPayloadHeader:

class MEVPayloadHeader(Container):
    payloadHeader: ExecutionPayloadHeaderV1
    feeRecipient: ExecutionAddress
    feeRecipientBalance: uint256

https://github.com/realbigsean/lighthouse/blob/mev-lighthouse/beacon_node/execution_layer/src/engine_api/json_structures.rs fills this out with some gaps, but none of it is directly specified and there are various points of ambiguity.

metachris:
I think we should use SSZ for encoding all the block/header fields in `builder_getHeaderV1` and `builder_getPayloadV1`, but not for the other regular params in the JSON-RPC. 

For instance, `setFeeRecipient` should send it's fields just as regular JSON-RPC params, and not SSZ encoded. 

ping @Ruteri @tersec @terencechain @lightclient

metachris:
Closing this issue now - it's in the spec with the reasoning as follows (cc/ @lightclient):

- We use SSZ encoding for the block data in the `builder_getPayloadV1 ` request
- The Engine API encodes everything as a JSON object, and to be consistent with that we use the same JSON objects in the responses as the EL client.
[...]

PR #87, opened 2022-04-08
spec: SSZ cleanup

metachris:
A proposal to clean up the SSZ confusion: use SSZ encoding for block/header data structures, but not for other JSON-RPC params (eg. `setFeeRecipient` params, or the `value` of field of the `getHeader` response).

I wonder if `ExecutionPayloadHeaderV1` and `ExecutionPayloadV1` in the responses shouldn't also be SSZ encoded?

See also #76 

Ping @lightclient @tersec @terencechain

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -56,22 +56,6 @@ Equivalent to `ExecutionPayloadV1`, except `transactions` is replaced with `tran
 
 Consider the following definitions supplementary to the definitions in [`consensus-specs`][consensus-specs].
 
-##### `builder_setFeeRecipientV1` Request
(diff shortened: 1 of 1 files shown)


lightclient:
@metachris I see that this is very confusing as it is written. The SSZ objects here are used only to compute the signing root that the message sender uses to compute their BLS signature. No SSZ objects are sent over JSON-RPC _except_ for the BlindBeaconBlock in `builder_getPayload`.

metachris:
Great, thanks for clearing this up, maybe we can just add a comment to the object

PR #91, opened 2022-04-14
A few updates to the spec

lightclient:
@mcdee shared some [good feedback](https://discord.com/channels/595666850260713488/874767108809031740/964048921049563176) on discord, so the following was updated:
* Renamed the SSZ containers to be more descriptive.
* Added `SignedBlindBeaconBlock` container.
* Encapsulated "messages" and "signatures" in the spec by creating objects out of what will be mapped into an SSZ for signing.
* Specified that `hash` in `GetHeader` refers to an execution layer block hash.
* Noted that the `feeRecipient` field of the header **MAY** deviate from the validator's mapped `feeRecipient` so long as there is a payment in the block to the expected `feeRecipient`.
* Removed a comment about malleability between the builder provided header and the signed blind beacon block from the validator.

Will open a separate PR to discuss removing `SetFeeRecipient`.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1,4 +1,4 @@
-# Version 0.3
+# Version 0.2.1
 
 This document specifies the Builder API methods that the Consensus Layer uses to interact with external block builders.
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -1,4 +1,4 @@
-# Version 0.2
+# Version 0.3
 
 This document specifies the Builder API methods that the Consensus Layer uses to interact with external block builders.
 
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -161,7 +161,7 @@ As `compute_signing_root` takes `SSZObject` as input, client software should con
 
 #### Specification
 1. Builder software **MUST** verify `signature` is valid under `publicKey`.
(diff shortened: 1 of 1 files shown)


metachris (approved):

docs/specification.md
```diff
@@ -1,4 +1,4 @@
-# Version 0.2
+# Version 0.3
```
metachris, line 1:
```suggestion
# Version 0.2
```

let's leave this version at 0.2 (if you really want to, increment to 0.2.1)

metachris (changes requested):
Please keep v0.2

docs/specification.md
lightclient: I just prefer to bump the version in some way when making backwards incompatible changes

metachris:
What's the reason to keep the setFeeRecipient timestamp requirement that it must not be from before the latest announcement?

metachris (approved):

lightclient:
@metachris how would you avoid messages getting replayed without that requirement?

metachris:
> @metachris how would you avoid messages getting replayed without that requirement?

Oh I see now... I was confused by the term announcent. It refers to the latest `setFeeRecipient` announcement, right?

PR #93, opened 2022-04-15
Use go-ethereum/rpc

metachris:
switch from gorilla/mux to go-ethereum/rpc


lib/rpcserver.go
```diff
@@ -0,0 +1,102 @@
...
+			Public:    true,
+		},
+	}
+	if err := node.RegisterApis(apis, []string{namespace}, srv, false); err != nil {
+		return nil, fmt.Errorf("could not register api: %s", err)
+	}
+	return srv, nil
+}
+
+// NewHTTPServer TODO
```
lightclient, line 79: Is this TODO for a comment or that this function needs work?

lib/service.go
```diff
@@ -29,17 +28,23 @@ type BoostService struct {
...
 }
 
-func newBoostService(relayURLs []string, log *logrus.Entry) (*BoostService, error) {
+func newBoostService(relayURLs []string, log *logrus.Entry, getHeaderTimeout time.Duration) (*BoostService, error) {
 	if len(relayURLs) == 0 || relayURLs[0] == "" {
 		return nil, errors.New("no relayURLs")
 	}
 
+	// Set custom GetHeader timeout
+	_getHeaderTimeout := defaultGetHeaderTimeout
```
lightclient, line 37: Seems like you could get rid of this assignment by directly setting `getHeaderTimeout` if it is `0`?
```go
if getHeaderTimeout == 0 {
    getHeaderTimeout = defaultGetHeaderTimeout
}
```

lightclient (changes requested):
LGTM

lib/service.go
```diff
@@ -29,17 +28,23 @@ type BoostService struct {
...
+	if getHeaderTimeout > 0 {
+		_getHeaderTimeout = getHeaderTimeout
+	}
+
 	return &BoostService{
 		relayURLs: relayURLs,
 		log:       log.WithField("prefix", "lib/service"),
 
 		httpClient:      http.Client{Timeout: defaultHTTPTimeout},
```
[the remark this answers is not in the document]
metachris: I don't think it's required. Why would it be?

lib/service.go
```diff
@@ -118,7 +119,7 @@ func (m *BoostService) SetFeeRecipientV1(_ *http.Request, args *[]interface{}, r
 			}
 
 			// Result should be true if any one relay responds true
-			*result = *result || _result
+			result = result || _result
```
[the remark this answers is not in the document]
metachris: Nope, they don't share a variable, the goroutines handle independent RPC requests and independent data.

lib/service.go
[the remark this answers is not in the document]
bhakiyakalimuthu: As per go net/http client, has to be a pointer. is there any reason ? :thinking: https://github.com/golang/go/blob/master/src/net/http/client.go#L586

lib/service.go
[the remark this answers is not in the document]
bhakiyakalimuthu: https://github.com/flashbots/mev-boost/pull/93/files#diff-c48208726be97e20698f1bd874d4bcece8e2af190e67fe6c1578aae3ba212bf4R93
Basically 'result` is a shared variable right ?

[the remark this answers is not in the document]
metachris: No, it's not shared. One instance per request.

lib/service.go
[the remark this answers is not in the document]
metachris: The line you like to does not mean that a client instance has to be referenced with a pointer. How your reference it does not matter at all for the instance itself.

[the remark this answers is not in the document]
metachris: The code you linked shows that the client methods are declared with pointer receivers, which means they can modify the underlying instance. https://go.dev/tour/methods/4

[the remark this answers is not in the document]
metachris: Accidentally applied this change due to fat fingers on the phone :-(

lib/service.go
```diff
@@ -81,22 +86,18 @@ type rpcResponseContainer struct {
 }
 
 // SetFeeRecipientV1 - returns true if at least one relay returns true
-func (m *BoostService) SetFeeRecipientV1(_ *http.Request, args *[]interface{}, result *bool) error {
+func (m *BoostService) SetFeeRecipientV1(ctx context.Context, feeRecipient, timestamp, publicKey, signature string) (*bool, error) {
```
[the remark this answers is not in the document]
metachris: used now, good catch 👍

lib/rpcserver.go
metachris: fixed, was just the comment to make the linter happy

lib/service.go
metachris: nice, done 🙏

lib/rpcserver.go
```diff
@@ -0,0 +1,102 @@
...
+	apis := []gethRpc.API{
+		{
+			Namespace: namespace,
+			Version:   "1.0",
+			Service:   backend,
+			Public:    true,
+		},
+	}
+	if err := node.RegisterApis(apis, []string{namespace}, srv, false); err != nil {
+		return nil, fmt.Errorf("could not register api: %s", err)
```
terencechain, line 74: Have we considered `%w` verb which supports wrapping?

```suggestion
		return nil, fmt.Errorf("could not register api: %w", err)
```

metachris: Right, makes sense! Will keep that in mind :)

lib/service.go
```diff
@@ -127,32 +127,33 @@ func (m *BoostService) SetFeeRecipientV1(_ *http.Request, args *[]interface{}, r
...
 	method := "builder_getHeaderV1"
 	logMethod := m.log.WithField("method", method)
 
 	if len(*blockHash) != 66 {
-		return fmt.Errorf("invalid block hash: %s", *blockHash)
+		return nil, fmt.Errorf("invalid block hash: %s", *blockHash)
 	}
 
 	// Call the relay
+	result := new(GetHeaderResponse)
```
jparyani, line 149:
```suggestion
	var result GetHeaderResponse
```
nitpick: unnecessary allocation, could be replaced by a `var` declaration

jparyani (approved):

lib/service.go
metachris: What would be the difference? `var` also allocates the memory 🤔

jparyani: This is a pointer variable, you’re allocating the value struct unnecessarily

jparyani: Ah my bad, the code suggestion missed the pointer, my bad :)

Issue #100, opened 2022-04-23
Crash in GetHeaderV1 method call


PtitLuca:
### Description

I was reading the code and testing the project when I realized that the `GetHeaderV1` method takes a pointer as parameter, the `blockHash`.

Here is the original function code :
```golang
// GetHeaderV1 TODO
func (m *BoostService) GetHeaderV1(ctx context.Context, blockHash *string) (*GetHeaderResponse, error) {
	method := "builder_getHeaderV1"
	logMethod := m.log.WithField("method", method)

	if len(*blockHash) != 66 {
		return nil, fmt.Errorf("invalid block hash: %s", *blockHash)
	}
        // The rest of the function
}
```

Dereferencing the pointer without pre-verification of its content (here a non-nil check) causes the method call to crash when providing a nil value for the `blockHash` parameter.

I know that the function is executed in a goroutine, but still, I think it impacts stability.

### Your environment

* **OS and version**: macOS Monterey v12.3.1
* **branch/commit hash that causes this issue**: 1108378fc37d9466f0bca96998c97062df195c04

### Steps to reproduce

* How to reproduce this issue ?<br />
As written in the project's README, I built mev-boost using the following command :
```shell
make build
```

And executed the binary using (don't mind the port, the default one was used by another program):
```shell
./mev-boost -port 8080
```

Then, in a new terminal, I ran the following curl command :
```shell
curl -X POST http://127.0.0.1:8080/ -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","method":"builder_getHeaderV1","params":[],"id":1}'
```

To confirm the behaviour, I also wrote the following test before patching the function :
```golang
func TestE2E_GetHeaderError(t *testing.T) {
	relay1 := setupMockRelay()
	server, err := newTestBoostRPCServer([]string{relay1.URL})
	require.Nil(t, err, err)
	defer server.Stop()

	client := gethRpc.DialInProc(server)
	defer client.Close()

	res := new(GetHeaderResponse)
	err = client.Call(&res, "builder_getHeaderV1", nil)
	require.Error(t, err)
}
```

And both of them lead to a goroutine crash.

* Where the issue is ?<br />
As mentionned in the description, the issue is caused by dereferencing a nil pointer.

### Expected behaviour
The server should reject my invalid request with an error message (I suppose ?).

### Actual behaviour
The goroutine handling the request crashed. Check out the logs for more info.

### Logs

With the CURL command :
```shell
[...]

PR #101, opened 2022-04-23
Add verification to prevent GetHeaderV1 method to crash

PtitLuca:
You can find more information about this PR in #100.


metachris:
Thanks!

PR #106, opened 2022-04-26
implement builder_status RPC call

metachris:


server/service.go
```diff
@@ -281,3 +284,8 @@ func (m *BoostService) GetPayloadV1(ctx context.Context, block string) (*types.E
 	}
 	return nil, fmt.Errorf("no valid GetPayloadV1 response from relay")
 }
+
+// Status implements the builder_status RPC method
+func (m *BoostService) Status(ctx context.Context) (*string, error) {
```
bhakiyakalimuthu, line 289: How about renaming `Status` to `Health` ? wdyt

bhakiyakalimuthu: Microservice/rest/k8s world uses this as Health 🤔

metachris: That seems to be not actually correct: https://kubernetes.io/docs/reference/using-api/health-checks/

> The healthz endpoint is deprecated (since Kubernetes v1.16), and you should use the more specific livez and readyz endpoints instead. 

Also the mev-boost spec uses `boost_status` already.

Thanks for the thoughts though.

PR #107, opened 2022-04-27
Spec updates

metachris:
Easiest to follow this PR if you read from bottom up (types -> service -> test -> spec)


server/service.go
```diff
@@ -21,7 +21,10 @@ var (
 )
 
 var (
-	errNoBlockHash = errors.New("no blockhash provided")
+	errInvalidHash      = errors.New("invalid blockhash provided")
```
Ruteri, line 24: Should the rpc errors implement error codes as per spec?

Ruteri (approved):

server/service.go
metachris: yes! will update codes in the next pr

types/types.go
```diff
@@ -87,32 +81,38 @@ type executionPayloadMarshallingOverrides struct {
...
-type SetFeeRecipientMessage struct {
+// RegisterValidatorRequestMessage as defined in https://github.com/flashbots/mev-boost/blob/main/docs/specification.md#request
+type RegisterValidatorRequestMessage struct {
 	FeeRecipient string `json:"feeRecipient"`
 	Timestamp    string `json:"timestamp"`
+	Pubkey       string `json:"pubkey"`
+}
+
+// IsValid returns true if the message properties are correct valid hex values
+func (r *RegisterValidatorRequestMessage) IsValid() bool {
```
Ruteri, line 92: I think this check is not necessary, as it's not really a validity check
I'd leave it non-validated until we have proper validation (parse `FeeRecipient` to `common.Address`, `Timestamp` to uint64, `Pubkey` to a proper pubkey
I think I'd also apply this to the rest of the validation - partial validation (in this case - making sure it's a hex) seems unnecessary

metachris: I will see if I can add types that better catch issues, but generally I think it's better to reject obviously wrong input, even if it's not perfect, rather than not checking at all.

server/service.go
metachris: See #108

types/types.go
metachris: Updated with proper types

Issue #116, opened 2022-05-02
Data races in GetHeaderV1 and RegisterValidatorV1


PtitLuca:
### Description

I was running the tests and having a look at the code in the `service.go` file while I discovered something strange.
It appeared that there was some pieces of the code where a data race could occur, so I decided to run the tests using the `-race` flag and I found a few of them.

### Your environment

* **OS and version**: macOS Monterey v12.3.1
* **branch/commit hash that causes this issue**: 79f11e0712e9e2b895133c4600d0b922cfbecebd

### Steps to reproduce

* How to reproduce this issue ?<br />
I ran the tests using the following command :
```shell
go test ./... -v -race
```

* Where the issue is ?<br />
There are actually several sub-issues causing the data races :
1. When spawning goroutines to query the relayers in both `GetHeaderV1` and `RegisterValidatorV1`, some values can be written at the same time by multiple goroutines
2. The mockserver does not have a data race protection for its `RequestCounter`

### Expected behaviour
The tests should pass without any data race.

### Actual behaviour
There were several data races. More below.

### Logs

```shell
?       github.com/flashbots/mev-boost/cmd/mev-boost    [no test files]
=== RUN   TestE2E_RegisterValidator
==================
WARNING: DATA RACE
Read at 0x00c0002eb63f by goroutine 21:
  github.com/flashbots/mev-boost/server.(*BoostService).RegisterValidatorV1.func1()
      /Users/ptitluca/GolandProjects/mev-boost/server/service.go:141 +0x7ac
  github.com/flashbots/mev-boost/server.(*BoostService).RegisterValidatorV1.func2()
      /Users/ptitluca/GolandProjects/mev-boost/server/service.go:142 +0x58

Previous write at 0x00c0002eb63f by goroutine 22:
  github.com/flashbots/mev-boost/server.(*BoostService).RegisterValidatorV1.func1()
      /Users/ptitluca/GolandProjects/mev-boost/server/service.go:141 +0x85c
  github.com/flashbots/mev-boost/server.(*BoostService).RegisterValidatorV1.func2()
      /Users/ptitluca/GolandProjects/mev-boost/server/service.go:142 +0x58

Goroutine 21 (running) created at:
  github.com/flashbots/mev-boost/server.(*BoostService).RegisterValidatorV1()
      /Users/ptitluca/GolandProjects/mev-boost/server/service.go:117 +0x6b4
  runtime.call128()
[...]

PR #117, opened 2022-05-02
Add data race protection in GetHeaderV1 and RegisterValidatorV1

PtitLuca:
You can learn more about the prupose of this PR in #116 .

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -130,6 +130,7 @@ func (m *BoostService) RegisterValidatorV1(ctx context.Context, message types.Re
 			}
 			if res.Error != nil {
 				logMethod.WithFields(logrus.Fields{"error": res.Error, "url": url}).Warn("error reply from relay")
+				// We can unlock using defer because we're leaving the routine right now
 				container.mu.Lock()
 				defer container.mu.Unlock()
 				container.lastRelayError = res.Error
@@ -175,9 +176,16 @@ func (m *BoostService) GetHeaderV1(ctx context.Context, slot hexutil.Uint64, pub
 		return nil, rpcErrInvalidPubkey
 	}
 
+	type safeHeaderContainer struct {
+		mu             sync.Mutex
+		result         *types.GetHeaderResponse
+		lastRelayError error
+	}
+	container := safeHeaderContainer{
+		result: new(types.GetHeaderResponse),
+	}
+
 	// Call the relay
-	result := new(types.GetHeaderResponse)
-	var lastRelayError error
 	var wg sync.WaitGroup
 	for _, relayURL := range m.relayURLs {
 		wg.Add(1)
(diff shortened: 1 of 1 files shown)


server/service.go
```diff
@@ -147,9 +157,9 @@ func (m *BoostService) RegisterValidatorV1(ctx context.Context, message types.Re
 
 	// If no relay responded true, return the last error message, or a generic error
 	var err error
-	if !ok {
-		err = lastRelayError
-		if lastRelayError == nil {
+	if !container.ok {
+		err = container.lastRelayError
+		if container.lastRelayError == nil {
```
terencechain, line 162: I think the `defer container.mu.Unlock()` at L150 will be executed on the completion of that goroutine... which means we probably want to hold the read lock here as well if we assume `RegisterValidatorV1` will be called concurrently

PtitLuca: Correct me if I'm wrong, but if `RegisterValidatorV1` is called by several goroutines in the JSON-RPC dispatcher, its internal values will not be shared across them ?

To simplify, I've written the following example :
```go
package main

import (
	"fmt"
	"sync"
)

func RegisterValidatorV1(wg *sync.WaitGroup, c chan bool) {
	defer wg.Done()

	type safeContainer struct {
		mu   sync.Mutex
		data bool
	}

	container := safeContainer{}

	var pool sync.WaitGroup

	for i := 0; i < 30; i++ {
		pool.Add(1)
		go func() {
			defer pool.Done()

			// Do some stuff, such as inverting container data value
			container.mu.Lock()
			defer container.mu.Unlock()

			container.data = !container.data
		}()
	}

	pool.Wait()

	<-c
	// Read container data
	fmt.Printf("container.data address: %p\n", &container.data)
}

func main() {
	var wg sync.WaitGroup
	defer wg.Wait()

	// We'll use this channel to start the goroutines at the same time
	c := make(chan bool)

	for i := 0; i < 50; i++ {
		wg.Add(1)
		go RegisterValidatorV1(&wg, c)
	}
	close(c)
}
```

Here, the `c` channel will ensure that the 50 initial goroutines will try to read `container.data` at the same time.
But as `container.data` is a local variable of the `RegisterValidatorV1`, I don't think there could be any data race here.

metachris: > Correct me if I'm wrong, but if RegisterValidatorV1 is called by several goroutines in the JSON-RPC dispatcher, its internal values will not be shared across them ?

That is correct.

server/service.go
```diff
@@ -109,8 +109,13 @@ func (m *BoostService) RegisterValidatorV1(ctx context.Context, message types.Re
...
 	}
 
-	ok := false // at least one builder has returned true
-	var lastRelayError error
+	type safeRegisterContainer struct {
+		mu             sync.Mutex
+		ok             bool // at least one builder has returned true
+		lastRelayError error
+	}
+	container := safeRegisterContainer{}
```
metachris, line 117: Why did you choose to create the container vs just adding the mutex?

PtitLuca: As the `ok` and `lastRelayError` fields both need to be protected by the mutex, I assumed it made more sense to combine them in a structure.
It shows that this mutex will "only" be used to protect these variables only.

Let me know if you disagree with this !

metachris: Makes sense :)

metachris:
Thanks for the contributions 👍

server/service.go
terencechain: That makes sense. Thanks for the explanations!

PR #123, opened 2022-05-15
Minor cleanups: return err, unused args...

terencechain:
Fixed some code warnings
 - rename `server` to `s` to not be the same as the import
 - remove unused arguments
 - return error for `RegisterName`


metachris:
Thanks! I hope you're aware that the current work is in the rest branch - PR #119. Expecting to finish the rest API and merge that on Monday.

PR #125, opened 2022-05-17
use flashbots/builder types

metachris:


0xpanoramix (approved):
Everything looks good 👍 !

README.md
```diff
@@ -85,22 +82,24 @@ and then run it with:
...
-git clone -b v021-upstream https://github.com/flashbots/mergemock.git
+git clone https://github.com/protolambda/mergemock.git
 cd mergemock
 go build . mergemock
 wget https://gist.githubusercontent.com/lightclient/799c727e826483a2804fc5013d0d3e3d/raw/2e8824fa8d9d9b040f351b86b75c66868fb9b115/genesis.json
+openssl rand -hex 32 | tr -d "\n" > jwt.hex
 ```
 
-Then you can run an integration test with mergemock, spawning both a mergemock execution engine and a mergemock consensus client as well as mev-boost:
+Then you can run an integration test with mergemock, spawning both a mergemock relay+execution engine and a mergemock consensus client pointing tomev-boost:
```
Ruteri, line 102: `to mev-boost`

server/service.go
```diff
@@ -50,7 +51,7 @@ func NewDefaultHTTPServerTimeouts() HTTPServerTimeouts {
 // BoostService TODO
 type BoostService struct {
 	listenAddr string
-	relays     []types.RelayEntry
+	relays     []boostTypes.RelayEntry
```
Ruteri, line 54: `RelayEntry` should be internal

0xpanoramix:
Just a minor comment, will you change the relay entry parsing in a separate PR ?
For now, it[ still uses the `hexutil.Bytes` from go-ethereum instead of the BLS from prysm](https://github.com/flashbots/mev-boost/blob/d86f680668692d2b3f093553520fa35cc16a9241/types/types.go#L9).

metachris:
> Just a minor comment, will you change the relay entry parsing in a separate PR ? For now, it[ still uses the `hexutil.Bytes` from go-ethereum instead of the BLS from prysm](https://github.com/flashbots/mev-boost/blob/d86f680668692d2b3f093553520fa35cc16a9241/types/types.go#L9).

Good question! I think we'll figure out the best approach once someone implements the getHeader response signature verification. I guess we should switch the types in this struct, yes

README.md
metachris: fixed

server/service.go
metachris: done 👍

Issue #126, opened 2022-05-18
Issue with Docker image based on scratch


metachris:
The scratch-based Docker image is not working.

```standard_init_linux.go:228: exec user process caused: no such file or directory```

* To run with scratch, it seems we need to disable CGO with `CGO_ENABLED=0`
* But with CGO disabled, part of the bls stack doesn't build

To reproduce the build issue: 

```
$ CGO_ENABLED=0 GOOS=linux go build -v -o mev-boost ./cmd/mev-boost

github.com/herumi/bls-eth-go-binary/bls
# github.com/herumi/bls-eth-go-binary/bls
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:9:28: undefined: SecretKey
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:9:40: undefined: Fr
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:13:26: undefined: Fr
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:13:31: undefined: SecretKey
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:19:28: undefined: PublicKey
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:19:40: undefined: G1
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:23:26: undefined: G1
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:23:31: undefined: PublicKey
../../../go/pkg/mod/github.com/herumi/bls-eth-go-binary@v0.0.0-20210917013441-d37c07cfda4e/bls/cast.go:29:23: undefined: Sign
[...]

PR #127, opened 2022-05-18
switch Docker image from scratch to alpine

metachris:
fixes https://github.com/flashbots/mev-boost/issues/126

Pushed to Docker hub: https://hub.docker.com/r/flashbots/mev-boost

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,9 @@ WORKDIR /build
 ADD . /build/
 RUN --mount=type=cache,target=/root/.cache/go-build make build-for-docker
 
-FROM scratch
+FROM golang:1.18
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,9 +4,9 @@ WORKDIR /build
 ADD . /build/
 RUN --mount=type=cache,target=/root/.cache/go-build make build-for-docker
 
-FROM scratch
+FROM golang:1.18
```
0xpanoramix, line 7:
```suggestion
FROM alpine

RUN apk add --no-cache libgcc libstdc++ libc6-compat
```

0xpanoramix (changes requested):
I've suggested changes according to the proposed solution in #126.

PR #128, opened 2022-05-21
use go-boost-utils, start relay check, registerValidator array

metachris:
cli cleanup, using types from go-boost-utils, and start of relay check at startup


server/service.go
```diff
@@ -50,7 +51,7 @@ func NewDefaultHTTPServerTimeouts() HTTPServerTimeouts {
 // BoostService TODO
 type BoostService struct {
 	listenAddr string
-	relays     []RelayEntry
+	relays     []*RelayEntry
```
Ruteri, line 54: Nit: why a pointer instead of a value? `RelayEntry` is a bunch of pointers and a string, it should be just fine as a value, and value semantics are usually preferable.

Ruteri (approved):

server/service.go
metachris: Good point, will update

PR #133, opened 2022-05-25
Only parse RelayEntry pubkey if it is present, adjust test data

Ruteri:
Redone all signatures after recent changes to marshaling.


metachris (approved):

metachris:
that's great, thanks!

PR #134, opened 2022-05-30
Adjust signing domain computation in test-cli

Ruteri:
Updated to match the spec:
* register validator is now signed with genesis fork version
* getHeader is now checked to be signed with bellatrix fork version

Reference: https://github.com/ethereum/builder-specs/blob/main/specs/builder.md#containers and https://github.com/ethereum/builder-specs/blob/main/specs/builder.md#signing

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/cmd/test-cli/main.go
+++ b/cmd/test-cli/main.go
@@ -41,7 +41,7 @@ func doRegisterValidator(v validatorPrivateData, boostEndpoint string, builderSi
 	log.WithError(err).Info("Registered validator")
 }
 
-func doGetHeader(v validatorPrivateData, boostEndpoint string, beaconNode Beacon, engineEndpoint string, proposerSigningDomain boostTypes.Domain) boostTypes.GetHeaderResponse {
+func doGetHeader(v validatorPrivateData, boostEndpoint string, beaconNode Beacon, engineEndpoint string, builderSigningDomain boostTypes.Domain) boostTypes.GetHeaderResponse {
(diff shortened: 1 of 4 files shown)


cmd/test-cli/main.go
```diff
@@ -186,28 +206,35 @@ func main() {
...
+		builderSigningDomain := computeDomain(boostTypes.DomainTypeAppBuilder, genesisForkVersionStr, boostTypes.Root{}.String())
+		proposerSigningDomain := computeDomain(boostTypes.DomainTypeBeaconProposer, bellatrixForkVersionStr, genesisValidatorsRootStr)
+		doGetPayload(mustLoadValidator(validatorDataFile), boostEndpoint, createBeacon(isMergemock, beaconEndpoint, engineEndpoint), engineEndpoint, builderSigningDomain, proposerSigningDomain)
 	default:
 		fmt.Println("Expected generate|register|getHeader|getPayload subcommand")
 		os.Exit(1)
 	}
 }
 
+func computeDomain(domainType boostTypes.DomainType, forkVersionHex string, genesisValidatorsRootHex string) boostTypes.Domain {
```
metachris, line 226: should this function live in `go-builder-utils`? probably not important but just wondering

Ruteri: Yeah, I don't think so since it's just a wrapper on internal options for usage with `boostTypes.ComputeDomain()`

PR #135, opened 2022-05-30
Update test framework

0xpanoramix:
**Related issue(s)**:
No related issue.

**Description**:
This PR adds various updates to the testing framework.

**Problem(s) & goal(s)**:
1. There's now a `testing` package which holds the logic related to the relay mocking and helper methods which are used in the tests.
2. The relay mock is now more generic : each relay holds its own key-pair and each of the response is signed properly by the associated key at compile-time.
3. Helper methods have been tested.
4. The relay entry method have been tested.
5. There's a new `backend` package, leaving the responsibility of implementing the `backend`'s interface to the `server` package. Regardless of the REST framework used to implement the server, `backend` holds a "common" interface for the backend. Some of the server values such as the endpoint paths are used by the testing framework.

**Additional comment**:
It is also possible to remove the `testing` and `backend` packages and keep the updates made to the test framework in the server package, but in several files to avoid confusion.
If the project is not aimed to be very large, this could be an option to consider as it is always possible to refactor this later on.

This could look like this :

```bash
├── server
│   ├── backend.go  // Holds the endpoints paths
│   ├── mock_relay.go // Holds the mocked relay
│   ├── mock_relay_test.go // Holds the mocked relay tests
│   ├── mock_types.go // Holds the helper methods
│   ├── mock_types_test.go // Holds the helper methods tests
│   ├── relay_entry.go
│   ├── relay_entry_test.go
│   ├── service.go
│   ├── service_test.go
│   ├── utils.go
│   └── utils_test.go
```

I'd really like to have your opinion about this :)

---

I have run these commands:

* [ ] `make lint`
* [ ] `make test`
* [ ] `go mod tidy`


metachris:
First off, thanks and great work!

I'm not against separating the testing but if there's no simple cli flag then perhaps better put it all together into the main package and keep the coverage high 🙂

metachris:
I like it a lot 👍

Issue #95, opened 2022-04-20
v1 implementation tasks


metachris:
Tasks:
* [x] #136 
* [ ] forwarding error codes from the relay (clarify error handling if 1/n or n/n relays fail, even with different results?)

wip:
* [x] use correct error payload format - #166 
* [ ] Store registerValidator payloads and resend to relays at regular intervals https://github.com/flashbots/mev-boost/pull/149

done:
* [x] verify signatures/payloads: #138 
* [x] require relay pubkey: #143 
* [x] check relays on startup: #140

PR #138, opened 2022-06-07
Add signature verifications

0xpanoramix:
**Description:**

This PR adds support for signature verifications in the following handlers :

- [x] `registerValidator`
- [x] `getHeader`
- [ ] `getPayload`

**Problem(s) & goal(s):**

See #95 for more context.

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

**Additional comments:**

For now, mergemock integration tests do not pass because the relay generates a random private key for each run and mev-boost does not have a way to know which one it is.
[I've created a PR](https://github.com/protolambda/mergemock/pull/51) which allows to provide a private key to mergemock's relay.

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -144,6 +144,16 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 			http.Error(w, errInvalidSignature.Error(), http.StatusBadRequest)
 			return
 		}
+
+		ok, err := types.VerifySignature(registration.Message, types.DomainBuilder, registration.Message.Pubkey[:], registration.Signature[:])
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		if !ok {
+			http.Error(w, errInvalidSignature.Error(), http.StatusBadRequest)
+			return
+		}
 	}
 
 	numSuccessRequestsToRelay := 0
--- a/server/service.go
+++ b/server/service.go
@@ -218,7 +218,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	var wg sync.WaitGroup
 	for _, relay := range m.relays {
 		wg.Add(1)
-		go func(relayAddr string) {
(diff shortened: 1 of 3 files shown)


server/service.go
```diff
@@ -144,6 +144,16 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 			http.Error(w, errInvalidSignature.Error(), http.StatusBadRequest)
 			return
 		}
+
+		ok, err := types.VerifySignature(registration.Message, types.DomainBuilder, registration.Message.Pubkey[:], registration.Signature[:])
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
metachris, line 150: we should probably log the error (with `log.Error`), for visibility for the operator

server/service.go
```diff
@@ -144,6 +144,16 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
+			return
+		}
+		if !ok {
+			http.Error(w, errInvalidSignature.Error(), http.StatusBadRequest)
```
metachris, line 154: same as above. maybe even wrap the error and add a bit of context 🤔

server/service.go
```diff
@@ -218,7 +228,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	var wg sync.WaitGroup
 	for _, relay := range m.relays {
 		wg.Add(1)
-		go func(relayAddr string) {
+		go func(relayAddr string, relayPubKey types.PublicKey) {
```
metachris, line 231: do we want to pass in the relay alltogether instead of the individual fields?

metachris:
very nice! 🙏

server/service.go
0xpanoramix: This way, we can only read the data. It slows down a little bit the process due to copies but honestly not _that_ much.

metachris (approved):

metachris:
merging into the develop branch

PR #140, opened 2022-06-08
Add relay check on startup

0xpanoramix:
**Related issue(s):**

For more information, see #95.

**Description:**

This PR adds a relay check whenever the server starts.
It queries each relay on their `status` endpoint.

For now, no behavior is associated to a missing relay.

**Additional context & references:**

I can add some logic when unavailable relays are found.
We can discuss about this below.

For now, if at least one relay is available, the method returns true. Otherwise if none are available, it returns false.

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -358,18 +358,15 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 
 // CheckRelays sends a request to each one of the relays previously registered to get their status
 func (m *BoostService) CheckRelays() bool {
-	var status = false
-
 	for _, relay := range m.relays {
 		m.log.WithField("relay", relay).Info("Checking relay")
 
 		err := SendHTTPRequest(context.Background(), m.httpClient, http.MethodGet, relay.Address+pathStatus, nil, nil)
 		if err != nil {
 			m.log.WithError(err).WithField("relay", relay).Error("relay check failed")
-		} else {
-			status = true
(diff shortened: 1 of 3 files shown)


server/service.go
```diff
@@ -355,3 +355,21 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
...
+// CheckRelays sends a request to each one of the relays previously registered to get their status
+func (m *BoostService) CheckRelays() bool {
+	var status = false
+
+	for _, relay := range m.relays {
+		m.log.WithField("relay", relay).Info("Checking relay")
+
+		err := SendHTTPRequest(context.Background(), m.httpClient, http.MethodGet, relay.Address+pathStatus, nil, nil)
+		if err != nil {
+			m.log.WithError(err).WithField("relay", relay)
```
metachris, line 368:
```suggestion
			m.log.WithError(err).WithField("relay", relay).Error("relay check failed")
```
should this case also return false right away?

metachris: I think we could define this behavior for `CheckRelays` until there is a need for more granular feedback about which relays work and which don't.

0xpanoramix: Got it, added in b51ac76b2aed3fe6645384af27c2e6a42de2c128

metachris:
Thanks!

PR #146, opened 2022-06-09
added docs for creating a pos privnet

metachris:
Credits go to @Ruteri


Ruteri:
I feel it's a bit too specific, maybe we should post the guide to github.com/flashbots/boost-geth-builder and link it here?

metachris:
I prefer this repository to be the main source of documentation and community, and think that having the docs here helps with that (instead of causing some fragmentation). The whole Builder FAQ and guides will also be in this repo.

Issue #141, opened 2022-06-08
Store validator preferences and resend them at a regular interval


0xpanoramix:
## Description

One of the features mentioned in #95 is to store the validator preferences (preferred fee recipient and gas limit for a given validator) and then send them at a regular interval to the relays.

As there are many ways to implement this feature, I think it could be interesting to have a discussion about which one is the best.

## 1 : Use a separate goroutine

We can have a goroutine which reads the registered preferences in a safe map and sends those values to the relays at a regular interval, something like :

```golang

// BoostService TODO
type BoostService struct {
	// Backend stuff...

	// The validators preferred fee recipient and gas limit
	mu          sync.RWMutex
	preferences map[string]types.SignedValidatorRegistration
}

// sendValidatorPreferencesAtInterval is launched as a goroutine
func (m *BoostService) sendValidatorPreferencesAtInterval(done chan bool) {
	ticker := time.NewTicker(time.Minute)
	defer ticker.Stop()

	for {
		select {
		case <-done:
			return

		case <-ticker.C:
			// Loads preferences from map after declaration
			payload := []types.SignedValidatorRegistration{}
			
			// Send to each relay
			for _, relay := range m.relays {
				err := SendHTTPRequest(context.Background(), m.httpClient, http.MethodPost, url, payload, nil)
			}
		}
	}
}

func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.Request) {
	// ...
	for _, registration := range payload {
		// Sanity checks...
		
		// Stores the SignedValidatorRegistration and maps it to the validator's public key
		m.mu.Lock()
		m.preferences[registration.Message.Pubkey.String()] = registration
		m.mu.Unlock()
	}
	// ...
}
```

I'd like to have your opinion about this, or if you have a better solution !

metachris:
That sounds like a good approach! Happy to go that direction 👍

0xpanoramix:
I have implemented this in #149, but I'm still wondering where to start the goroutine.

I was thinking about [here](https://github.com/0xpanoramix/mev-boost/blob/fa367969d833ce57c63c5c94e3b3abc7992d25a9/server/service.go#L112), what's your opinion on this ?

And what should be the name of a new CLI flag used to provide this interval, if we need one at all (as it could be a hardcoded value).

cc @metachris

metachris:
That place to start the goroutine looks good 👍 Maybe rename the function to `startServer`.

[...]

PR #149, opened 2022-06-10
Add storage and resend for validators preferences

0xpanoramix:
**Description:**

This PR adds a way to send validators preferences at a regular interval to the relays.
It follows the implementation proposed in #141 and is part of the steps to achieve v1 as described in #95.

**Problem(s) & goal(s):**

//

**Additional context & references:**

This PR adds a method to handle the logic of sending the preferences to the relays, which is used by both the standard handler and a new method used to send the already registered prefs at a provided interval.

Several questions still remains regarding the implementation, see #141 for more.

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`


metachris:
Very nice, and thoughtful! I noticed that `registerValidatorAtInterval` is not yet used, but it's still marked as wip so I'm sure you're aware ^^

cmd/mev-boost/main.go
```diff
@@ -21,16 +21,18 @@ var (
...
 	// defaults
-	defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
-	defaultRelayTimeoutMs     = getEnvInt("RELAY_TIMEOUT_MS", 2000) // timeout for all the requests to the relay
-	defaultRelayCheck         = os.Getenv("RELAY_STARTUP_CHECK") != ""
-	defaultGenesisForkVersion = getEnv("GENESIS_FORK_VERSION", "")
+	defaultListenAddr                = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
+	defaultRelayTimeoutMs            = getEnvInt("RELAY_TIMEOUT_MS", 2000) // timeout for all the requests to the relay
+	defaultRelayCheck                = os.Getenv("RELAY_STARTUP_CHECK") != ""
+	defaultGenesisForkVersion        = getEnv("GENESIS_FORK_VERSION", "")
+	defaultRegisterValidatorInterval = getEnvInt("REGISTER_VALIDATOR_INTERVAL_SEC", 5)
```
Ruteri, line 28: Let's think about a sane default, if all of the validators send their registration every 5 seconds it will effectively be a DDoS attack on the unfortunate relay on the receiving end
Slot is 12 seconds, that's a sane lower bound
An epoch is 384 seconds, and validators are asked to re-register once per epoch so it's an upper bound
I think something like a fourth (96s) of an epoch is already frequent enough

server/service.go
```diff
@@ -129,6 +146,60 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
+	log := m.log.WithField("method", "registerValidatorAtInterval")
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-done:
+			return
+		case <-ticker.C:
```
Ruteri, line 196: One issue that arises is that we can send twice in a short interval (if a registration request comes near the timer tick)
In this approach it's not easy to solve it, and I think we could come at this from a different angle:
* make registerValidatorAtInterval the only place responsible for sending the registration
* listen on both the timer, and a new channel that accepts new registration requests
* on new registration request, send the preferences and reset the timer

This has some nice benefits:
* m.vp is now an object local to registerValidatorAtInterval
* validator preferences struct can become just a plain old map
* no need for mutex
* after a registration you will correctly wait until the next interval

The only downside is that you'll need to return the registration status back over a channel (or a callback), but it should be okay.

We can submit a PR to this PR if you prefer not implementing it

cmd/mev-boost/main.go
metachris: even resending once an epoch should be fine (and instantly on receiving)

server/service.go
0xpanoramix: Totally agree with this comment ! Thanks a lot for raising the issue and proposing an alternative 👍
I can make the updates don't worry.

cmd/mev-boost/main.go
0xpanoramix: I agree with your comments, but it seems like it's a bad idea to let the user decide of the interval in the end, wdyt ?

server/service.go
0xpanoramix: Added in de04d2f038dcde4efcd2bc8b7928911a762db457

cc @Ruteri

server/service.go
```diff
@@ -164,32 +237,10 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
-			numSuccessRequestsToRelay++
-		}(relay.Address)
-	}
-
-	// Wait for all requests to complete...
-	wg.Wait()
+	// Send the payload to the goroutine responsible for handling the resend at interval
+	m.newRegistrationsRequests <- payload
+	// Block until we get the number of successful requests back from this goroutine
+	numSuccessRequestsToRelay := <-m.numSuccessRequestsToRelay
```
Ruteri, line 243: Once context (timeout) handling is implemented this will become a race, the result channel (or callback) should be specific to the request.
I think pushing a pair of (payload, result chan) would be a clean way to solve it that will also work nicely with context timeouts.

server/service.go
```diff
@@ -56,10 +57,17 @@ type BoostService struct {
 	serverTimeouts       HTTPServerTimeouts
 
 	httpClient http.Client
+
+	// Used to stop registerValidatorAtInterval
+	done chan bool
+	// Used by registerValidator to share new incoming registration request with the goroutine holding the ticker
```
Ruteri, line 63: It'd be great to avoid shared fields as they usually introduce a race

server/service.go
```diff
@@ -56,10 +57,17 @@ type BoostService struct {
 	serverTimeouts       HTTPServerTimeouts
 
 	httpClient http.Client
+
+	// Used to stop registerValidatorAtInterval
+	done chan bool
```
Ruteri, line 62: Another way to implement shutdowns will be to have a context with cancel, and storing the cancel function.

server/service.go
```diff
@@ -110,10 +122,15 @@ func (m *BoostService) StartHTTPServer() error {
 		IdleTimeout:       m.serverTimeouts.Idle,
 	}
 
+	// Start separate process to send validator preferences at regular interval.
+	go m.registerValidatorAtInterval(time.Second*384, m.done)
```
Ruteri, line 126: I'm guessing this should be registerValidatorIntervalSec

server/service.go
```diff
@@ -384,3 +435,7 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 		return
 	}
 }
+
+func (m *BoostService) shutdown() {
+	m.done <- true
```
metachris, line 440: using a context would be better because if there's multiple things to shut down, this channel would only work for one of them. could be okay for now, and we can also do this in another PR.

server/service.go
```diff
@@ -129,6 +146,62 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-done:
+			// mev-boost has probably stopped
+			return
+		case payload = <-m.newRegistrationsRequests:
```
metachris, line 194: this will overwrite the previous payload, but it's conceivable that there's multiple validators using one BN that would then overwrite each others registrations. perhaps better to make this additive?

Ruteri: I'm actually not sure about it, the expectation is to have a single BN connecting to mev-boost, and forcing this to be additive would mean you cannot ever remove a registration.

server/service.go
0xpanoramix: I agree, that's what they've been made after all.
Fixed in a2bb054510f72b4545958aa409ad48ffe7700382

server/service.go
0xpanoramix: Missed that, thanks ! Fixed in e1d657247437af9b985129d2e6a1ec7f1b8891c6.

server/service.go
metachris: @Ruteri expectation is a single BN, but this single BN could have _many_ validator clients connected! We can not ignore this case.

One approach could be to add a time to each registration and remove old ones.

server/service.go
```diff
@@ -129,6 +147,62 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-m.ctx.Done():
+			// mev-boost has probably stopped
+			return
+		case srr := <-m.registrationsRequests:
+			// registerValidator has received new registrations and forwards them to here
+			srr.numSuccessRequestsToRelay <- m.sendValidatorPreferences(log, srr.preferences)
```
metachris, line 197: `payload` isn't updated

0xpanoramix: 👍 
Fixed in a3d5a4b578ce5c133f749112f71ba096230488f4

metachris:
It was closed because the temporary develop branch was deleted after merging into main. Please point that PR towards the main branch, I seem to not be able to do it from my user, and can't reopen until then :)

PR #150, opened 2022-06-13
Implement a RelayManager based on mev-boost

alrevuelta:
## Description
This PR implements a RelayManager using `mev-boost` as a baseline. A `RelayManager` is a multiplexer of relays that also monitors, updates, and uses the reputation of each relay before connecting to it. The main goal is to never request a `signedHeader` from a relay that in the past failed to return a `getPayload` or that returned a wrong one. The reputation also benefits relays that respond faster.

Note that the use of the `RelayManager` introduces a change in the architecture. With this approach, `mev-boost` no longer runs in the proposer. A single `RelayManager` entity is run by a trusted party, and every proposer that trusts it connects via [`http-mev-relay`](https://github.com/prysmaticlabs/prysm/blob/v2.1.3-rc.3/cmd/beacon-chain/flags/base.go#L15). The `RelayManager` also knows the relays endpoints out there, and serves as a router between the proposer needing a block and a set of relays/builders willing to facilitate it, always providing the most reputed one based on past responses.

Note that this approach has clear some clear pros/cons:
* Con: Single point of failure
* Pro: Simplicity
* Pro: Proposers do not need to run extra software

## Architecture
image

## Reputation

See document for more details

## TODOs
- [x] Implement a better reputation algorithm

---

I have run these commands:

* [ ] `make lint`
* [x] `make test`
* [x] `go mod tidy`


metachris:
Thanks for the suggestions and interesting conversations in DMs. Closing this now.

PR #153, opened 2022-06-16
Add storage and resend for validators preferences

0xpanoramix:
## Description

This PR adds a way to send validators preferences at a regular interval to the relays.
It follows the implementation proposed in #141 and is part of the steps to achieve v1 as described in #95.

## Context & references

You can have a look at #149 for old comments about the implementation.

## Additional comments

This PR adds a method to handle the logic of sending the preferences to the relays, which is used by both the standard handler and a new method used to send the already registered prefs at a provided interval.

---

## I have run these commands

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`


metachris:
Thanks a lot for the contribution! As discussed in DMs, we're not yet sure whether it'd preferable to delegate this task to the validator client. Leaving the PR open in the meantime.

metachris:
closing this PR for now, let's reopen if the need for this functionality arises, but I'd be happy to leave the logic up to the validators for the sake of simplicity.

PR #166, opened 2022-06-22
Use correct error format

avalonche:
Use correct http error format in mev-boost as part of #95


server/service_test.go
```diff
@@ -43,7 +44,7 @@ func newTestBackend(t *testing.T, numRelays int, relayTimeout time.Duration) *te
 	return &backend
 }
 
-func (be *testBackend) request(t *testing.T, method string, path string, payload any) *httptest.ResponseRecorder {
+func (be *testBackend) request(t *testing.T, method string, path string, payload interface{}) *httptest.ResponseRecorder {
```
metachris, line 47: leave that at `any` (supported since go 1.18 iirc)

server/service_test.go
```diff
@@ -200,7 +216,7 @@ func TestGetHeader(t *testing.T) {
 		"0x8a1d7b8dd64e0aafe7ea7b6c95065c9364cf99d38470c12ee807d55f7de1529ad29ce2c422e0b65e3d5a05c02caca249")
 	path := getPath(1, hash, pubkey)
 	require.Equal(t, "/eth/v1/builder/header/1/0xe28385e7bd68df656cd0042b74b69c3104b5356ed1f20eb69f1f925df47a3ab7/0x8a1d7b8dd64e0aafe7ea7b6c95065c9364cf99d38470c12ee807d55f7de1529ad29ce2c422e0b65e3d5a05c02caca249", path)
-
+
```
metachris, line 219: perhaps set your IDE to automatically trim trailing spaces: https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

server/service.go
```diff
@@ -21,12 +21,26 @@ var (
 	errInvalidHash      = errors.New("invalid hash")
 	errInvalidPubkey    = errors.New("invalid pubkey")
 	errInvalidSignature = errors.New("invalid signature")
+	errRelayResponse    = errors.New("no successful relay response")
```
metachris, line 24: maybe a bit better name would be `errNoSuccessfulRelayResponse`

metachris:
thanks, this is great!

PR #171, opened 2022-06-24
Simplify OK responses across handlers

0xpanoramix:
## Description
This PR adds a new method, `respondOK`, which avoids code duplication across the mev-boost server's handlers.

## Context & references
It's similar to the `respondError` method, but returns a 200 Status code and an optional element instead.

## Additional comments

---

## I have run these commands

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -27,6 +27,7 @@ var (
 )
 
 var nilHash = types.Hash{}
+var nilResponse = struct{}{}
 
 type httpErrorResp struct {
 	Code    int    `json:"code"`
@@ -42,18 +43,13 @@ func respondError(w http.ResponseWriter, code int, message string) {
 }
 
 func respondOK(w http.ResponseWriter, result any) error {
-	var err error
-
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
 
-	if result == nil {
-		_, err = fmt.Fprintf(w, `{}`)
-	} else {
-		err = json.NewEncoder(w).Encode(result)
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		return err
 	}
-
-	return err
+	return nil
 }
 
 // BoostService TODO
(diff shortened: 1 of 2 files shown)


server/service.go
```diff
@@ -41,6 +41,21 @@ func respondError(w http.ResponseWriter, code int, message string) {
 	}
 }
 
+func respondOK(w http.ResponseWriter, result any) error {
+	var err error
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	if result == nil {
```
avalonche, line 50: instead of passing `nil` you can using an empty struct like `var nilResponse struct{}` and pass `nilResponse` into `respondOK()`. The empty struct will be encoded as `{}\n`, making this function a bit cleaner:
```
if err := json.NewEncoder(w).Encode(result); err != nil {
    return err
}
return nil
```

avalonche (approved):

server/service.go
0xpanoramix: Is it okay to have this `\n` ? If yes, I can make this change.

metachris: yes, let's just make the proposed change, and get this merged 👍

0xpanoramix: Alright, added in cff01182a553449187c63d803abf0241520c1bb6

server/service.go
```diff
@@ -41,6 +42,16 @@ func respondError(w http.ResponseWriter, code int, message string) {
...
 }
 
+func respondOK(w http.ResponseWriter, result any) error {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		return err
+	}
+	return nil
```
metachris, line 52:
```suggestion
	return json.NewEncoder(w).Encode(result)
```

could simplify like this

metachris:
thanks!

PR #172, opened 2022-06-24
Add relays statuses calls in boost status endpoint

### LINKED ISSUE #169: Reflect relays statuses in `status` endpoint
I think `mev-boost` status endpoint should call `status` endpoints for each relays and return OK if at least one returned OK. Otherwise it should return KO (503?).

0xpanoramix:
## Description
This PR updates the status handler, by calling each relay's status endpoint.

## Context & references
This PR implements and resolves #169.

## Additional comments
I think we should wait for :
- `inputs from the other CL clients`
- Applying #171 changes 

Before merging this PR.

---

## I have run these commands

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`


server/service.go
```diff
@@ -116,10 +116,24 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
 	fmt.Fprintf(w, `{}`)
 }
 
+// handleStatus sends calls to the status endpoint of every relay.
+// It returns OK if at least one returned OK, and returns KO otherwise.
 func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
-	fmt.Fprintf(w, `{}`)
+	for _, relay := range m.relays {
```
avalonche, line 122: You can parallelise the call to check relay status. See https://github.com/flashbots/mev-boost/blob/main/server/service.go#L160-L193 for an example of how it's done.

0xpanoramix: Thanks ! Added in 9d7cab19f8b96cefe73e15e821c516c863ae7457

server/service.go
```diff
@@ -116,10 +117,41 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
...
+// It returns OK if at least one returned OK, and returns KO otherwise.
 func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
-	fmt.Fprintf(w, `{}`)
+	var wg sync.WaitGroup
+	var numSuccessRequestsToRelay uint32
+
+	// At the end, we wait for every routine and return status according to relay's ones.
+	defer func() {
```
metachris, line 127: i'm not a big fan of this defer because of the indirection -- does it provide any advantage over putting the code just after the range loop? this would feel simpler to me intuitively 🤔

server/service.go
```diff
@@ -116,10 +117,41 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
...
+		}
+	}()
+
+	for _, r := range m.relays {
+		wg.Add(1)
+
+		go func(relay RelayEntry) {
+			defer wg.Done()
+
+			m.log.WithField("relay", relay).Info("Checking relay status")
```
metachris, line 145:
```suggestion
			m.log.WithField("relay", relay).Debug("Checking relay status")
```

server/service.go
0xpanoramix: I agree, let's not add spaghetti code while it's possible ^^

server/service.go
```diff
@@ -116,10 +117,41 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
...
+
+			url := relay.Address + pathStatus
+			err := SendHTTPRequest(context.Background(), m.httpClient, http.MethodGet, url, nil, nil)
+
+			if err == nil {
```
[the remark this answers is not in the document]
0xpanoramix: Changed in 71e94ff2b278a610a092146275d3e68bccb7f5a2

metachris (approved):

server/service.go
```diff
@@ -121,10 +122,39 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
...
+	var wg sync.WaitGroup
+	var numSuccessRequestsToRelay uint32
+
+	for _, r := range m.relays {
+		wg.Add(1)
+
+		go func(relay RelayEntry) {
+			defer wg.Done()
+
+			m.log.WithField("relay", relay).Debug("Checking relay status")
```
metachris, line 137:
```suggestion
			m.log.WithField("relay", relay.Address).Debug("Checking relay status")
```

metachris: maybe better would be to do something like

```suggestion
			log = log.WithField("relay", relay)
			log.Debug("Checking relay status")
```

and later down on the error logging also use the new log method

server/service.go
```diff
@@ -126,10 +127,40 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
...
+			defer wg.Done()
+
+			log := m.log.WithField("relay", relay.Address)
+			log.Debug("Checking relay status")
+
+			url := relay.Address + pathStatus
+			err := SendHTTPRequest(context.Background(), m.httpClient, http.MethodGet, url, nil, nil)
+
+			if err != nil {
+				m.log.WithError(err).Error("failed to retrieve relay status")
```
metachris, line 149:
```suggestion
				log.WithError(err).Error("failed to retrieve relay status")
```
```suggestion
				m.log.WithError(err).Error("failed to retrieve relay status")
```

metachris:
Thanks!

PR #182, opened 2022-06-30
Add verifications in getHeader handler

### LINKED ISSUE #174: Verify parentHash on getHeader response
Verification of the parentHash would be good to add to the getHeader relay response handler. Are there other attributes to verify too?

0xpanoramix:
## 📝 Summary

This PR adds the following verification(s) in the `getHeader` handler :
- Checks if the `parent_hash` in the response (given by the relay) is the same as the one in the request (the one given by the proposer)
- The same way, checks if the `public_key` in the response is the same as the one in the request

It closes #174.

## ⛱ Motivation and Context

More context about this PR can be found in #174.

## 📚 References

The `getHeader` handler is defined [here](https://ethereum.github.io/builder-specs/#/Builder/getHeader)

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -281,7 +281,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			}
 
 			// Verify response coherence with proposer's input data
-			if responsePayload.Data != nil && responsePayload.Data.Message.Header.ParentHash.String() != parentHashHex {
+			if responsePayload.Data.Message.Header.ParentHash.String() != parentHashHex {
 				log.WithError(errInvalidParentHash).Error("proposer and relay parent hashes are not the same")
 				return
 			}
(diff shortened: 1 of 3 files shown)


server/service.go
```diff
@@ -279,6 +280,12 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 				return
 			}
 
+			// Verify response coherence with proposer's input data
+			if responsePayload.Data != nil && responsePayload.Data.Message.Header.ParentHash.String() != parentHashHex {
```
metachris, line 284: `responsePayload.Data` nil check is already in line 261

```suggestion
			if responsePayload.Data.Message.Header.ParentHash.String() != parentHashHex {
```

server/service.go
```diff
@@ -279,6 +280,12 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
+				log.WithError(errInvalidParentHash).Error("proposer and relay parent hashes are not the same")
```
metachris, line 285: maybe could add the relevant fields to the message (`log.WithFields`)

metachris:
great, thanks!

metachris:
ty

Issue #154, opened 2022-06-17
Support validator account level relay choice


beetrootkid:
As a node operator, different customers want to select different sources of MEV. To enable efficient distribution of validator keys across regions and consensus client types, our validator clients hold keys from different customers.

Teku and Lighthouse support specification of fee_recipient at the validator key level.

It would be great if mev-boost / builder-api could utilize the same file but with an added parameter per validator key to select the appropriate relay when that particular key is chosen to propose a block.

This would also enable solo stakers who have more than one validator to explore different sources of MEV without having to run multiple nodes.

jmcruz1983:
Could we extend `relay urls` argument to support file base configuration? 
https://github.com/flashbots/mev-boost/issues/155

metachris:
Could you provide a specific example?

An issue is that mev-boost and the relay needs to conform to the builder spec: https://ethereum.github.io/builder-specs/

This API spec only allows setting `fee_recipient` and `gas_limit` for a specific validator, and mev-boost is just proxying that data to the relays which prepare it for the builders. There is no way for mev-boost to select additional preferences at the builders.

beetrootkid:
Ah - so, should I raise this at the builder-spec repo?

A specific example:

- One customer "trusts" relays 3,4,5 as sources of MEV
- Another customer "trusts" relays 1,2,3

Right now, the only way to handle this is to have different consensus clients for the two customers with differently associated MEV boost sidecars...

metachris:
hmmm i see, i think this would indeed be easiest handled in mev-boost. and it seems related to #155, file based relay list, where you could set this relation. we'll discuss internally and report back in a little bit.

metachris:
For reference, the Teku config file format looks like this:

```json
{
  "proposer_config": {
    "0xa057816155ad77931185101128655c0191bd0214c201ca48ed887f6c4c6adf334070efcd75140eada5ac83a92506dd7a": {
      "fee_recipient": "0x50155530FCE8a85ec7055A5F8b2bE214B3DaeFd3",
      "validator_registration": {
        "enabled": true,
        "gas_limit": "12345654321"
      }
    }
  },
  "default_config": {
    "fee_recipient": "0x6e35733c5af9B61374A128e6F85f553aF09ff89A",
    "validator_registration": {
      "enabled": false
    }
  }
}
```

metachris:
The most simple config for mev-boost could look like this (and allow per-proposer overrides): ```json
{
  "default_config": {
[...]

PR #186, opened 2022-07-05
Add validator relay selection from configuration file

0xpanoramix:
## 📝 Summary

This PR adds support for loading a configuration file used by mev-boost to understand which validator trusts which relay.
The configuration file currently follows this format (this can change in the future):

```typescript
interface configuration {
    builder_relays_groups: {
        [key: string]: string[];
    };
    proposer_config: {
        [key: string]: {
            fee_recipient: string;
            builder_registration: {
                enabled: boolean;
                builder_relays: string[];
                gas_limit: string;
            }
        };
    },
    default_config: {
        fee_recipient: string;
        builder_registration: {
            enabled: boolean;
            builder_relays: string[];
        }
    }
}
```

Any input on implementation is welcomed !
But for consistency purposes, any high-level ideas should go to #154.

Goals:
- [x] Support configuration extraction from raw JSON content.
- [x] Add support default configuration extraction.
- [x] Support configuration extraction from a JSON file.
- [x] Determine the way to store proposer configurations.
- [x] Add sanity checks for configuration values.
    - [x] Relay endpoints must be valid.
    - [x] Reference to groups must be valid.
- [x] Add a new CLI flag to provide the configuration filename.
- [x] Add a new field in the `Boost Configuration` struct.
- [x] Use the proper relays in the handlers.
- [x] Write tests to cover these cases.

Optional:
- [ ] Groups' names must match a pattern / are not empty.
- [ ] Remove duplicates after extraction.
- [ ] What should the default configuration file look like ?

## ⛱ Motivation and Context

For more context, see #154.

## 📚 References

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`


server/config.go
```diff
@@ -0,0 +1,175 @@
...
+type configuration struct {
+	BuilderRelaysGroups map[string][]string          `json:"builder_relays_groups"`
+	ProposerConfig      map[string]rawProposerConfig `json:"proposer_config"`
+	DefaultConfig       rawProposerConfig            `json:"default_config"`
+}
+
+// configFromJSON builds a new configuration from a given JSON raw object.
+func configFromJSON(raw json.RawMessage) (*configuration, error) {
+	config := &configuration{
+		BuilderRelaysGroups: make(map[string][]string),
```
metachris, line 32: are the `make` here needed?

cmd/mev-boost/main.go
```diff
@@ -41,6 +41,8 @@ var (
 	relayTimeoutMs = flag.Int("request-timeout", defaultRelayTimeoutMs, "timeout for requests to a relay [ms]")
 	relayCheck     = flag.Bool("relay-check", defaultRelayCheck, "check relay status on startup and on the status API call")
 
+	configPath = flag.String("config", "", "configuration for proposer preferences")
```
metachris, line 44:
```suggestion
	configPath = flag.String("config", "", "configuration file for proposer preferences")
```

configuration.json
```diff
@@ -0,0 +1,17 @@
+{
```
metachris, line 1: should this be in gitignore?

server/config.go
```diff
@@ -0,0 +1,185 @@
+package server
+
+import (
+	"encoding/json"
+	"errors"
+	"github.com/flashbots/go-boost-utils/types"
+	"io/ioutil"
+)
+
+type rawConfiguration struct {
```
metachris, line 10: perhaps `configurationEntry` could be a better name? 🤔

server/config.go
```diff
@@ -0,0 +1,185 @@
...
+	placeholder := &rawConfigurationFile{}
+	if err := json.Unmarshal(bytes, placeholder); err != nil {
+		return nil, err
+	}
+
+	return placeholder, nil
+}
+
+// ConfigurationStorage holds one proposer configuration.
+type ConfigurationStorage struct {
```
metachris, line 45: maybe `ProposerRelayConfig` could be a better name

server/config.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	configuration := &ConfigurationStorage{
+		FeeRecipient: feeRecipient,
+		Enabled:      rawConf.ValidatorRegistration.Enabled,
+		GasLimit:     gasLimit,
+	}
+
+	if len(rawConf.ValidatorRegistration.BuilderRelays) == 0 {
+		return nil, errors.New("no builder is associated to this proposer")
```
metachris, line 128: this might actually be allowed, ping @tbenr

server/service.go
```diff
@@ -210,21 +214,25 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
 	ua := UserAgent(req.Header.Get("User-Agent"))
 
-	for _, relay := range m.relays {
-		wg.Add(1)
-		go func(relay RelayEntry) {
-			defer wg.Done()
-			url := relay.GetURI(pathRegisterValidator)
-			log := log.WithField("url", url)
+	for _, registration := range payload {
+		configurationStorage := m.pcs.GetProposerConfiguration(registration.Message.Pubkey)
```
metachris, line 218:
```suggestion
		proposerConfig := m.pcs.GetProposerConfiguration(registration.Message.Pubkey)
```

server/service.go
```diff
@@ -242,12 +250,18 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	vars := mux.Vars(req)
 	slot := vars["slot"]
 	parentHashHex := vars["parent_hash"]
-	pubkey := vars["pubkey"]
+
+	publicKey, err := types.HexToPubkey(vars["pubkey"])
+	if err != nil || len(vars["pubkey"]) != 98 {
```
metachris, line 255:
```suggestion
	if err != nil {
```

size is already checked in HexToPubkey

metachris:
This looks great! 👍 

Left a few minor comments, and needs a rebase. I think we might move the logic into it's own directory, like `/proposerconfig`, to keep things cleanly separated, and it's enough code to split up the entry and storage into separate files.

server/proposer_pk_storage.go
```diff
@@ -0,0 +1,9 @@
+package server
+
+import "github.com/flashbots/go-boost-utils/types"
+
+// proposerPublicKeyStorage is used to store a proposer's public key according to a specific slot
+// along with the block hash it is currently proposing.
+type proposerPublicKeyStorage struct {
+	storage map[string]types.PublicKey
```
metachris, line 8: instead of a concatenated string as key, maybe use a struct with explicit slot and hash.

also a helper method on this to expire entries older than a slot or two might be useful.

server/service.go
0xpanoramix: My bad, I forgot to create an issue about this but I added this check to prevent `TestGetHeader/Invalid pubkey length` to fail.
If you remove this condition you'll see this test fail.

cmd/mev-boost/main.go
0xpanoramix: Updated in 3eaa39a9fe2dc59163cfdd0d5c496a606396f4f8, I could not apply your suggestion due to rebase.

configuration.json
0xpanoramix: I think it's better to have a default one already there, but renaming it is a good idea. Done in e38e46812c6005656c62152a213db00f960750ed.

server/config.go
tbenr: If it refers to a specific pubkey, yes I think it should be optional. 
The idea is that default config must have the relay set defined and specific pubkey config can optionally override it. 
cc @StefanBratanov

server/config.go
0xpanoramix: Renamed in ce4e6dcd160154120f6bdabd57751cc445913b6d.

server/service.go
0xpanoramix: I've [opened a PR](https://github.com/flashbots/go-boost-utils/pull/26) which fixes the issue in go-boost-utils.

server/proposer_pk_storage.go
0xpanoramix: Added in 7f917d8ef552501817ba72ec602baea5fe16e6c4 and 63f5716a80628e7e28662829651241468ef3543a.

server/service.go
0xpanoramix: Fixed in db51af85f88f43ca0bb9398c78abfbc409fc67cc.

jmcruz1983:
Any update here?

ricardolyn:
it will be nice to have this ready for testing long before Mainnet as we need to prepare our staking platform for it. thank you :)

metachris:
Started some simplifications here: https://github.com/flashbots/mev-boost/commit/00c4f8ad9df5121715bbd600fb7de490e8ba4565

A test fails and carrying the proposer pubkey into getPayload has an issue, but that's the general direction i'd like to move. Trying to simplify things a little.

Sadly can't open a PR against this PR because it's in another repo where I don't have write access.

server/config.go
metachris: cc @0xpanoramix -- shouldn't this just fall back to the default config here rather than return an error?

0xpanoramix: It definitely should, good catch thanks !

server/service.go
```diff
@@ -229,23 +229,27 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
 	})
 
-	relayRespCh := make(chan error, len(m.relays))
+	relayRespCh := make(chan error, len(m.pcs.GetAllRelays()))
 
-	for _, relay := range m.relays {
-		go func(relay RelayEntry) {
-			url := relay.GetURI(pathRegisterValidator)
-			log := log.WithField("url", url)
+	for _, registration := range payload {
```
metachris, line 234: this loop logic seems wrong. 

now its doing: for each payload, send the full payload to each relay that proposer enabled.

let's say 10 registrations and using 2 relays -- this will send all the registrations 10 times to both relays 🤔

server/service.go
```diff
@@ -259,46 +263,51 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
 	var mu sync.Mutex
 	relays := make(map[string][]string) // relays per blockHash
 	result := bidResp{}
 
 	ua := UserAgent(req.Header.Get("User-Agent"))
 
+	result.proposerPublicKeys = append(result.proposerPublicKeys, publicKey)
+
+	// Select the relays we want mev-boost to reach.
+	configurationStorage := m.pcs.GetProposerConfiguration(publicKey)
```
metachris, line 302:
```suggestion
	proposerConfig := m.pcs.GetProposerConfiguration(publicKey)
```

server/service.go
```diff
@@ -259,46 +263,51 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
+	})
+	log.Debug("getHeader")
+
 	var mu sync.Mutex
 	relays := make(map[string][]string) // relays per blockHash
 	result := bidResp{}
 
 	ua := UserAgent(req.Header.Get("User-Agent"))
 
+	result.proposerPublicKeys = append(result.proposerPublicKeys, publicKey)
```
metachris, line 299: why a slice? there'll only be a single public key 🤔

metachris, line 299: shouldn't there be only a single proposer per slot?

server/service.go
```diff
@@ -431,7 +445,33 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
...
 	defer requestCtxCancel()
 
-	for _, relay := range m.relays {
+	key := bidRespKey{
+		slot:      payload.Message.Slot,
+		blockHash: payload.Message.Body.ExecutionPayloadHeader.BlockHash.String(),
+	}
+
+	// It means that someone is trying to call getPayload without having called getHeader first.
+	fmt.Printf("KEY %+v\n", key)
```
metachris, line 454: remove debug leftovers :)

server/service.go
```diff
@@ -431,7 +445,33 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
...
+		// TODO: Log error.
+		// log.WithField("", "").Errorf("")
+		m.respondError(w, http.StatusBadGateway, errNoSuccessfulRelayResponse.Error())
+		return
+	}
+	publicKeys := m.bids[key].proposerPublicKeys
+	publicKey := types.PublicKey{}
+
+	for _, publicKey = range publicKeys {
+		ok, _ := types.VerifySignature(payload.Message, m.builderSigningDomain, publicKey[:], payload.Signature[:])
```
metachris, line 466: that signature check doesn't fail on invalid signature. i think we can leave that to the relay too, we don't check the signatures in other places in mev-boost either.

server/service.go
```diff
@@ -431,7 +445,33 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
...
+		blockHash: payload.Message.Body.ExecutionPayloadHeader.BlockHash.String(),
+	}
+
+	// It means that someone is trying to call getPayload without having called getHeader first.
+	fmt.Printf("KEY %+v\n", key)
+	fmt.Printf("BIDS %+v\n", m.bids[key])
+	if m.bids[key] == nil {
+		// TODO: Log error.
+		// log.WithField("", "").Errorf("")
+		m.respondError(w, http.StatusBadGateway, errNoSuccessfulRelayResponse.Error())
```
metachris, line 459: maybe it should use all relays if no bid was found 🤔

server/service.go
0xpanoramix: The issue is that I need this check to get the public key associated to the current proposer.

1. I want to get the current proposer's public key in order to call its relays (based on the configuration file).
2. Because this value is not given as a parameter in the `getPayload` call, I use the list of public keys made from the various calls to `getHeader` associated to this `slot + blockchash` (bidKey). One of these public keys is the proposer's one, the only way to know which one it is is to try to verify the signature of the `getPayload` parameter : if it indeed verifies the signature, it's the public key I'm looking for. Otherwise it's an another proposer's public key.

server/service.go
0xpanoramix: Fixed in e6c020a55472ee55c4bc47ed2f0d2c58231610ec.
I'm sorry I could not commit your suggestion as I am using this variable later on and I needed to rename it there as well.

server/service.go
0xpanoramix: Thanks ! Removed in 4cad6ecf18a1454f0ce34f95a1a9bb642bf0fd11.

server/service.go
0xpanoramix: Alright, removed in eb6cbc1543f2f572b298ec6c663f9c34263d7793.

server/service.go
0xpanoramix: Yep it was a misunderstood on my side. Fixed in eb6cbc1543f2f572b298ec6c663f9c34263d7793.

server/config.go
```diff
@@ -0,0 +1,185 @@
...
+
+		configurationStorage, err := newConfigurationStorage(&configuration, raw.BuilderRelaysGroups)
+		if err != nil {
+			return nil, err
+		}
+
+		pcs.proposerConfigurations[address] = configurationStorage
+	}
+
+	return pcs, nil
```
ricardolyn, line 92: How can we dynamically update the configuration without restarting MEV-Boost to force the configuration file to be reread?

sergeyWh1te: Same question.
For example, https://github.com/spf13/viper provides out of the box dynamically updates config files without restarting goLang app.

metachris:
After consulting with @0xpanoramix we're closing this for now, and may revisit this feature again later on.

metachris:
This feature adds extensive complexity around the choice of relays, and is impacting the getPayload calls significantly.

To get this feature over the line, a first step would be to refactor mev-boost to move common functionality into a `common` package, and then extract the `config` part into it's own package.

This PR gets a lot of things right, and much of it could be reused to implement this feature eventually.

If someone wants to pick this up, the effort would be supported.

PR #190, opened 2022-07-06
getHeader 204 (no content)

metachris:
## 📝 Summary

* Handle 204 response from the relay
* Return 204 on getHeader requests where no relay has a Bid

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/server/service_test.go
+++ b/server/service_test.go
@@ -257,8 +257,8 @@ func TestGetHeader(t *testing.T) {
 		rr = backend.request(t, http.MethodGet, path, nil)
 		require.Equal(t, 2, backend.relays[0].GetRequestCount(path))
 		require.Equal(t, 2, backend.relays[1].GetRequestCount(path))
-		require.Equal(t, `{"code":502,"message":"no successful relay response"}`+"\n", rr.Body.String())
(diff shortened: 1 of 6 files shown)


server/service_test.go
```diff
@@ -257,8 +257,8 @@ func TestGetHeader(t *testing.T) {
 		rr = backend.request(t, http.MethodGet, path, nil)
 		require.Equal(t, 2, backend.relays[0].GetRequestCount(path))
 		require.Equal(t, 2, backend.relays[1].GetRequestCount(path))
-		require.Equal(t, `{"code":502,"message":"no successful relay response"}`+"\n", rr.Body.String())
-		require.Equal(t, http.StatusBadGateway, rr.Code, rr.Body.String())
+		require.Equal(t, http.StatusNoContent, rr.Code)
+		// require.Equal(t, `{"code":502,"message":"no successful relay response"}`+"\n", rr.Body.String())
```
terencechain, line 261: Should we remove these commented codes? do we plan to uncomment them later? If we do, perhaps a TODO or some mentioning would be nice

metachris: yes, i'll remove!

PR #191, opened 2022-07-06
Fix some things I noticed when reviewing

jtraglia:
## 📝 Summary

* Fix broken link to installing section.
* Add pipe (`|`) to table header line, technically invalid without it.
* Change "schema" -> "scheme", common mistake.
* Fix a few problems in privnet instructions; noteably add a missing word.
* Fix a few minor problems in comments.
* Remove misleading/contradicting comment about finding best fee.

## ⛱ Motivation and Context

I can't help but fix small things when I see them.

## 📚 References

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,9 +9,9 @@ Please start by reading our [code of conduct](CODE_OF_CONDUCT.md).
 Install a few dev dependencies:
 
 ```bash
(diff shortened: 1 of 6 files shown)


metachris:
That's great, thanks!

PR #194, opened 2022-07-06
Cleanup makefile

jtraglia:
## 📝 Summary

* Add `all` and `clean` targets.
  * I'm sure I'm not the only one that's tried just running `make` to build things.
  * Using `git clean -fdx` instead of manually specifying files as we already depend on `git` for the version.
* Set fake targets (happens to be all of them) as phony (see reference).
  * Not really that important, but I think it's good practice.
* Delete `run` and `run-dev` rules. I don't think these make sense anymore.
  * `mev-boost` has required options (network) now.
* Delete the `generate-ssz` rule. This has been moved to `go-boost-utils` (see reference).
  * Took me a few minutes to realize that the `types` directory moved repos.
* Set dependencies for docker rules. Should be able to just do `docker-push`.
  * Didn't actually test this part, so please double-check.

## ⛱ Motivation and Context

I like Makefiles and I wanted to clean things up a bit.

## 📚 References

* https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
* https://github.com/flashbots/go-boost-utils/blob/417a64679a5d4a2c12b64ff7ee918243d4ab34eb/Makefile#L23-L25

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,11 +33,6 @@ lint:
 	go vet ./...
 	staticcheck ./...
 
-.PHONY: generate-ssz
-generate-ssz:
-	rm -f types/builder_encoding.go
-	sszgen --path types --include ../go-ethereum/common/hexutil --objs Eth1Data,BeaconBlockHeader,SignedBeaconBlockHeader,ProposerSlashing,Checkpoint,AttestationData,IndexedAttestation,AttesterSlashing,Attestation,Deposit,VoluntaryExit,SyncAggregate,ExecutionPayloadHeader,VersionedExecutionPayloadHeader,BlindedBeaconBlockBody,BlindedBeaconBlock,RegisterValidatorRequestMessage,BuilderBid,SignedBuilderBid
-
 .PHONY: test-coverage
 test-coverage:
 	go test -race -v -covermode=atomic -coverprofile=coverage.out ./...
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,73 +4,87 @@ MERGEMOCK_BIN=./mergemock
...
 run-mergemock-integration: build
 	make -j3 run-boost-with-relay run-mergemock-consensus run-mergemock-relay
 
+.PHONY: build-for-docker
 build-for-docker:
 	GOOS=linux go build -ldflags "-X main.version=${VERSION}" -v -o mev-boost ./cmd/mev-boost
 
-docker-image:
+.PHONY: docker-image
+docker-image: build-for-docker
```
metachris, line 78:
```suggestion
docker-image:
```

`build-for-docker` is executed inside the build step (part of `docker build`)

Makefile
```diff
@@ -4,73 +4,87 @@ MERGEMOCK_BIN=./mergemock
...
 	DOCKER_BUILDKIT=1 docker build . -t mev-boost
 	docker tag mev-boost:latest ${DOCKER_REPO}:${VERSION}
 	docker tag mev-boost:latest ${DOCKER_REPO}:latest
 
-docker-push:
+.PHONY: docker-push
+docker-push: docker-image
```
metachris, line 84:
```suggestion
docker-push:
```

i prefer to have building and pushing of the docker images decoupled. we could even remove this, since the images are pushed to docker by CI nowadays.

metachris:
Thanks a lot, these are great! 👍 

Left 2 suggestions, and since merging the other PR this needs to be rebased :)

jtraglia:
Ah you're right. Thanks for pointing that out. Please review again.

metachris:
very nice, thank you!

PR #206, opened 2022-07-12
Default MaxHeaderSize to 4kb

avalonche:
## 📝 Summary

Add flag for boost to configure max header size, defaulting to 4kb.

## ⛱ Motivation and Context

This is to prevent DoS attacks by preventing attackers sending oversized headers to mev-boost. 

## 📚 References

---

* [ ] `make lint`
* [ ] `make test`
* [ ] `go mod tidy`


metachris:
Please remove the cli flag for it, I'd prefer not accumulating Flags for every setting. If we want to make it configurable then we can start with only an env var, but i don't expect this to change much.

PR #209, opened 2022-07-14
Remove redundant signature check

terencechain:
Payload's signature is already default at 96 bytes due to `Signature` type, `handleGetPayload`'s signature length check is redundant and can be removed

```go
package types

type Signature [96]byte
```


avalonche (approved):

metachris:
good catch, thanks!

PR #212, opened 2022-07-15
server read + readheader timeout

metachris:
## 📝 Summary

Read and readheader timeout. 

---

* [x] `make lint`
* [x] `make test`
* [x] `go mod tidy`


server/service.go
```diff
@@ -131,11 +131,12 @@ func (m *BoostService) StartHTTPServer() error {
 		Addr:    m.listenAddr,
 		Handler: m.getRouter(),
 
-		ReadTimeout:       0,
-		ReadHeaderTimeout: 0,
+		ReadTimeout:       600 * time.Millisecond,
```
metachris, line 134: may be too low, in case it's over the network? 🤔

.github/workflows/releaser.yaml
```diff
@@ -52,5 +52,5 @@ jobs:
         with:
           tag_name: ${{ github.ref }}
           release_name: ${{ github.ref }}
-          draft: false
+          draft: true
```
avalonche, line 55: what is this changed for?

server/service.go
avalonche: maybe you can make it configurable as an env var?

metachris: done, please recheck

avalonche (approved):

server/service.go
```diff
@@ -131,11 +132,12 @@ func (m *BoostService) StartHTTPServer() error {
...
-		ReadHeaderTimeout: 0,
-		WriteTimeout:      0,
-		IdleTimeout:       0,
-		MaxHeaderBytes:    m.maxHeaderBytes,
+		ReadTimeout:       time.Duration(config.ServerReadTimeoutMs) * time.Millisecond,
+		ReadHeaderTimeout: time.Duration(config.ServerReadHeaderTimeoutMs) * time.Millisecond,
+		WriteTimeout:      time.Duration(config.ServerWriteTimeoutMs) * time.Millisecond,
+		IdleTimeout:       time.Duration(config.ServerIdleTimeoutMs) * time.Millisecond,
+
+		MaxHeaderBytes: m.maxHeaderBytes,
```
avalonche, line 140: max header bytes can also be a config var

.github/workflows/releaser.yaml
metachris: by default the release has no good description but is sent out with email nevertheless. better to start as draft, manually update the description and then release

server/service.go
metachris: +1

PR #216, opened 2022-07-16
print usage when run without args (fails because it needs a network)

metachris:
## 📝 Summary

If it fails when run without args, print the usage too

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -33,7 +33,7 @@ var (
 	// cli flags
 	printVersion = flag.Bool("version", false, "only print version")
 	logJSON      = flag.Bool("json", defaultLogJSON, "log in JSON format instead of text")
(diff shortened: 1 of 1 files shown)


cli/main.go
```diff
@@ -33,7 +33,7 @@ var (
 	// cli flags
 	printVersion = flag.Bool("version", false, "only print version")
 	logJSON      = flag.Bool("json", defaultLogJSON, "log in JSON format instead of text")
-	logLevel     = flag.String("loglevel", defaultLogLevel, "log-level: trace, debug, info, warn/warning, error, fatal, panic")
+	logLevel     = flag.String("loglevel", defaultLogLevel, "minimum loglevel: trace, debug, info, warn/warning, error, fatal, panic")
```
avalonche, line 36: why is the description about minimum log level?

metachris: because this flag sets the minimum loglevel

PR #229, opened 2022-08-03
if no payload is received, log which relays sent the corresponding bid

metachris:
## 📝 Summary

When sending a signedBlindedBeaconBlock to all the relays, and no payload is received, mev-boost should log which relay the bid was coming from, to know which relay was causing the problem. For this, we need to remember the bid, to recover it in getPayload.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -22,7 +22,6 @@ var (
 	errInvalidSlot               = errors.New("invalid slot")
 	errInvalidHash               = errors.New("invalid hash")
 	errInvalidPubkey             = errors.New("invalid pubkey")
-	errInvalidSignature          = errors.New("invalid signature")
 	errNoSuccessfulRelayResponse = errors.New("no successful relay response")
 
 	errServerAlreadyRunning = errors.New("server already running")
@@ -46,7 +45,7 @@ type BoostServiceOpts struct {
 	RelayCheck            bool
 }
 
-// BoostService TODO
+// BoostService - the mev-boost service
 type BoostService struct {
 	listenAddr string
 	relays     []RelayEntry
@@ -56,6 +55,9 @@ type BoostService struct {
 
 	builderSigningDomain types.Domain
 	httpClient           http.Client
+
+	bidsLock sync.Mutex
+	bids     map[bidRespKey]bidResp // keeping track of bids, to log the originating relay on withholding
 }
 
 // NewBoostService created a new BoostService
@@ -74,6 +76,7 @@ func NewBoostService(opts BoostServiceOpts) (*BoostService, error) {
 		relays:     opts.Relays,
 		log:        opts.Log.WithField("module", "service"),
 		relayCheck: opts.RelayCheck,
+		bids:       make(map[bidRespKey]bidResp),
 
 		builderSigningDomain: builderSigningDomain,
 		httpClient: http.Client{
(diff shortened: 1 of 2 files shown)


server/service.go
```diff
@@ -326,31 +346,49 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
+			result.t = time.Now()
 		}(relay)
 	}
 
 	// Wait for all requests to complete...
 	wg.Wait()
 
-	if result.Data == nil || result.Data.Message == nil || result.Data.Message.Header == nil || result.Data.Message.Header.BlockHash == nilHash {
-		log.Info("no bids received from relay")
+	if result.response.Data == nil {
```
metachris, line 364: the other fields must be present because they are checked in line 313 already

server/service.go
```diff
@@ -416,8 +454,13 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 	// Wait for all requests to complete...
 	wg.Wait()
 
+	// If no payload has been received from relay, log loudly about withholding!
 	if result.Data == nil || result.Data.BlockHash == nilHash {
-		log.Warn("getPayload: no valid response from relay")
+		bidKey := bidRespKey{slot: payload.Message.Slot, blockHash: payload.Message.Body.ExecutionPayloadHeader.BlockHash.String()}
+		m.bidsLock.Lock()
```
Ruteri, line 460: Q: can we respond before locking and logging?

server/service.go
```diff
@@ -326,31 +346,49 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
+	}).Info("best bid")
+
+	// Remember the bid, for future logging in case of withholding
+	bidKey := bidRespKey{slot: _slot, blockHash: result.response.Data.Message.Header.BlockHash.String()}
+	m.bidsLock.Lock()
+	m.bids[bidKey] = result
+	m.bidsLock.Unlock()
+
+	// Return the bid
+	m.respondOK(w, result.response)
```
Ruteri, line 386: since more than one relay can respond with the same (slot, blockhash) it's probably better to save all relays responses

server/service.go
metachris: yes!

server/service.go
```diff
@@ -345,45 +348,53 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
 
-	if result.response.Data == nil {
+	if result.blockHash == "" {
 		log.Info("no bid received")
 		w.WriteHeader(http.StatusNoContent)
 		return
 	}
 
+	// Return the bid
+	m.respondOK(w, result.response)
```
Ruteri, line 381: responding to getHeader before updating `m.bids` has the drawback of a possible race with the subsequent getPayload
I don't think this is likely to happen in the wild, since the proposer has to sign the request but it's not out of the question

metachris: indeed, perhaps better to keep it at the end

server/service.go
Ruteri: this is sadly only going to work for getPayload, not getHeader

Ruteri (approved):

PR #232, opened 2022-08-03
Check bid key exists before logging

terencechain:
Would it be possible for the caller sends a block hash/slot combo that doesn't exist in the `bids` map, and relay also responded nil for result? In this case, I think `originalResp.relays` will result in a panic


metachris:
Thanks! But shouldn't be an issue, because if not found then it would return nil, which `strings.Join` can handle: https://go.dev/play/p/V1KzwgD_0eZ

metachris:
please comment again if i'm missing something!

thanks for the contribution terence, as always 🙏

PR #235, opened 2022-08-04
Adjust mergemock integration test cleanup

Ruteri:
## 📝 Summary

Current implementation won't work if there's a problem with any of the intermediate PID values, as the for loop won't be ran.

## ⛱ Motivation and Context

Killing based on jobs -p should work in most cases.

## 📚 References

---

* [ ] `make lint`
* [ ] `make test-race`
* [ ] `go mod tidy`


metachris (approved):

metachris:
thanks!

PR #238, opened 2022-08-05
Use blockhash as tiebreaker for equal bids

metachris:
## 📝 Summary

Uses blockhash as tiebreaker for equal bids. See also https://github.com/ethereum/builder-specs/issues/25

## ⛱ Motivation and Context

Previously it was first come first serve

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -357,9 +357,17 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 				relays[blockHash] = append(relays[blockHash], relay.String())
 			}
 
-			// Skip if value (fee) is not greater than the current highest value
-			if result.response.Data != nil && responsePayload.Data.Message.Value.Cmp(&result.response.Data.Message.Value) < 1 {
-				return
+			// Compare the bid with already known top bid (if any)
+			if result.response.Data != nil {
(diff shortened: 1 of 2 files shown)


server/service.go
```diff
@@ -357,9 +357,17 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 				relays[blockHash] = append(relays[blockHash], relay.String())
 			}
 
-			// Skip if value (fee) is not greater than the current highest value
-			if result.response.Data != nil && responsePayload.Data.Message.Value.Cmp(&result.response.Data.Message.Value) < 1 {
-				return
+			// Compare the bid with already known top bid (if any)
+			if result.response.Data != nil {
```
terencechain, line 361: I think it'll be great to add unit tests for these core additions. Perhaps refactor into its function to make it more testable as well. Also happy to help on it after the PR is merged if we want to merge it sooner

metachris: I'd like that! Please take a stab :)

terencechain (dismissed):

metachris:
Perhaps random would be more fair than blockhash.

PR #239, opened 2022-08-05
getHeader: verify that pubkey in message is the relay-key we trust

metachris:
## 📝 Summary

Verifies that the public key in the bid matches the pre-configured relay public key. Might be necessary to avoid issues with BLS signatures.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


Ruteri (approved):

ra--:
looks good! :+1:

PR #252, opened 2022-08-18
Parse slot/gas as uint64 in test code to prevent downcasting

jtraglia:
## 📝 Summary

* Parse `slot` and `gas` with `ParseUint` instead of `Atoi`.

## ⛱ Motivation and Context

The values were being parsed as signed integers and converted to unsigned 64-bit integers.

This was identified with Semgrep & the `string-to-int-signedness-cast` rule.

image

## 📚 References

* https://semgrep.dev/
* https://semgrep.dev/r?q=trailofbits.go.string-to-int-signedness-cast.string-to-int-signedness-cast
* https://github.com/trailofbits/semgrep-rules/blob/main/go/string-to-int-signedness-cast.yml

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cmd/test-cli/beacon.go
+++ b/cmd/test-cli/beacon.go
@@ -58,12 +58,12 @@ func getCurrentBeaconBlock(beaconEndpoint string) (beaconBlockData, error) {
 		return beaconBlockData{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


metachris:
thanks!

Issue #257, opened 2022-08-22
Empty block selection


nick-staked:
MEV-Boost: v0.7.10

We've noticed empty flashbots blocks on Sepolia. Ex: https://sepolia.etherscan.io/block/1716295

```json
{
  "blockHash": "0x8c911e603021c4a7e632f3768411ceeac65be25c0a69734c0c5f0eac7b618e07",
  "blockNumber": 1716295,
  "level": "info",
  "method": "getHeader",
  "module": "service",
  "msg": "best bid",
  "parentHash": "0xaf37d6b28480e30ed6010d1f194eafe617b61fdb5bdc71bedf75549d7501fd38",
  "pubkey": "0x875977457a3a801e2a25d728bd3424535d82abc9d473d785b6a66b66d9bbac5ff66166ae6ae16485fa2e326828100373",
  "relays": "https://0x845bd072b7cd566f02faeb0a4033ce9399e42839ced64e8b2adcfc859ed1e8e1a5a293336a49feac6d9a5edb779be53a@builder-relay-sepolia.flashbots.net?id=staked",
  "slot": "426081",
  "time": "2022-08-18T18:16:13Z",
  "txRoot": "0x7ffe241ea60187fdb0187bfa22de35d1f9bed7ab061d9401fd47e34a54fbede1",
  "value": "0"
}
```

In cases like these, where the best bid is `0`, is it preferable to default back to the local EE or use the flashbots block?

metachris:
Thanks for the report. We've now disabled sending empty blocks on the testnets.

PR #261, opened 2022-08-23
ignore bids with 0 value

metachris:
## 📝 Summary

Don't use blocks without value

See also #257

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


ralexstokes:
what do you think about *also* checking if the execution payload header has a transaction root equal to the root of the empty list?

if *either* condition holds then mev-boost filters that bid

metachris:
Great suggestion with the tx root, added.

metachris:
Yeah, so mergemock returns bids without transactions and the mergemock tests fail now :)

Time to deprecate mergemock?

Issue #273, opened 2022-08-28
Allow setting a minimum bid value to determine whether or not to use a block supplied by the builder API


allboxes:
Currently users are required to make a binary choice between using a centralized block builder or forgoing MEV.

MEV-Boost should allow user's to set a fixed fee (e.g. 0.2 ETH) under which they will use a locally produced block.

In this way validators can gain exposure to large MEV opportunities while still contributing to decentralized block building when MEV is small/non existent.

Additionally, as the CL does not currently have a view into the value of a locally produced block, this provides some protection to user's from relay's producing very low value blocks.

In the future this can be enhanced by comparing the value of the locally built block vs. the remote builder's submission and using a fee delta instead of a fixed fee.

Originally submitted as an issue to a few CLs but it was suggested that it would be better to implement it in MEV-Boost.
https://github.com/ConsenSys/teku/issues/6140
https://github.com/sigp/lighthouse/issues/3517

allboxes:
I took a stab at the PR, I have minimal GO experience but it seems to work fine.

allboxes:
@metachris any thoughts on this?

metachris:
Thanks for the contribution! And sorry for being slow to respond on this. It was on the radar, but due to the merge there were so many other priorities.

I wonder if there's an appetite for this feature from a wider audience. Could people please signify their interest by 👍 the initial comment above?

allboxes:
> @allboxes, I think this PR is step in the right direction given that many people are wary of running mev-boost especially since the mev-boost reward has the possibility of being less than the reward you would have earned from a locally produced block. I have 2 comments (although this may be beyond the scope of the PR)

>    There are 2 types of min bid, absolute min bid and relative min bid. I take it this PR is an absolute mid bid, but it would also be useful to have the option to specify a relative min bid which would be the min bid above the reward you would receive from a locally produced block

>   Is there a way of transmitting this information to the relay? Under this implementation if I have a min bid of .1 eth will the relay realize this and only submit bids of .1 eth or more? Will the bid immediately be rejected, giving the relay the opportunity to submit a higher bid? Or will the low bid be silently ignored? The last case seems sub-optimal if the relay is willing to bid higher but did not know of the min required bid of the validator.

@jumboshrimp100, responding here to keep the PR clean.

[...]

PR #274, opened 2022-08-28
Min bid

allboxes:
## 📝 Summary

Allow users to set a minimum bid to accept a block from a relay

## ⛱ Motivation and Context

https://github.com/flashbots/mev-boost/issues/273

---

* [✅ ] `make lint`
* [✅] `make test-race`
* [✅] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -2,11 +2,13 @@ package cli
 
 import (
 	"flag"
+	"math/big"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/flashbots/go-boost-utils/types"
 	"github.com/flashbots/mev-boost/config"
 	"github.com/flashbots/mev-boost/server"
 	"github.com/sirupsen/logrus"
@@ -25,6 +27,7 @@ var (
 	defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
 	defaultRelayCheck         = os.Getenv("RELAY_STARTUP_CHECK") != ""
 	defaultGenesisForkVersion = getEnv("GENESIS_FORK_VERSION", "")
+	defaultRelayMinBidEth     = getEnvFloat64("RELAY_MIN_BID", 0.001)
 	defaultDisableLogVersion  = os.Getenv("DISABLE_LOG_VERSION") == "1" // disables adding the version to every log entry
 
 	// mev-boost relay request timeouts (see also https://github.com/flashbots/mev-boost/issues/287)
@@ -46,6 +49,7 @@ var (
 	listenAddr       = flag.String("addr", defaultListenAddr, "listen-address for mev-boost server")
 	relayURLs        = flag.String("relays", "", "relay urls - single entry or comma-separated list (scheme://pubkey@host)")
 	relayCheck       = flag.Bool("relay-check", defaultRelayCheck, "check relay status on startup and on the status API call")
+	relayMinBidEth   = flag.Float64("min-bid", defaultRelayMinBidEth, "minimum bid to accept from a relay [eth]")
 	relayMonitorURLs = flag.String("relay-monitors", "", "relay monitor urls - single entry or comma-separated list (scheme://host)")
 
 	relayTimeoutMsGetHeader  = flag.Int("request-timeout-getheader", defaultTimeoutMsGetHeader, "timeout for getHeader requests to the relay [ms]")
@@ -161,6 +165,19 @@ func Main() {
 		}
 	}
 
+	if *relayMinBidEth < 0.0 {
+		log.Fatal("Please specify a non-negative minimum bid")
+	}
+
+	if *relayMinBidEth > 1000000.0 {
+		log.Fatal("Minimum bid is too large, please ensure min-bid is denominated in Ethers")
+	}
+
+	relayMinBidWei, err := floatEthTo256Wei(*relayMinBidEth)
+	if err != nil {
+		log.WithError(err).Fatal("failed converting min bid")
+	}
+
 	opts := server.BoostServiceOpts{
 		Log:                      log,
 		ListenAddr:               *listenAddr,
@@ -168,6 +185,7 @@ func Main() {
 		RelayMonitors:            relayMonitors,
 		GenesisForkVersionHex:    genesisForkVersionHex,
 		RelayCheck:               *relayCheck,
+		RelayMinBid:              *relayMinBidWei,
 		RequestTimeoutGetHeader:  time.Duration(*relayTimeoutMsGetHeader) * time.Millisecond,
 		RequestTimeoutGetPayload: time.Duration(*relayTimeoutMsGetPayload) * time.Millisecond,
 		RequestTimeoutRegVal:     time.Duration(*relayTimeoutMsRegVal) * time.Millisecond,
@@ -200,4 +218,32 @@ func getEnvInt(key string, defaultValue int) int {
 		}
 	}
 	return defaultValue
+}
+
+func getEnvFloat64(key string, defaultValue float64) float64 {
+	if value, ok := os.LookupEnv(key); ok {
+		val, err := strconv.ParseFloat(value, 64)
+		if err == nil {
+			return val
+		}
+	}
+	return defaultValue
+}
+
+func floatEthTo256Wei(val float64) (*types.U256Str, error) {
+	bigval := new(big.Float)
+	bigval.SetFloat64(val)
+
+	wad := new(big.Float)
+	wad.SetInt(big.NewInt(1000000000000000000))
+
+	bigval.Mul(bigval, wad)
+
+	result := new(big.Int)
+	bigval.Int(result)
+
+	u256 := new(types.U256Str)
+	err := u256.FromBig(result)
+
+	return u256, err
 }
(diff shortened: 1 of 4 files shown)


server/service.go
```diff
@@ -369,6 +372,11 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 				relays[blockHash] = append(relays[blockHash], relay.String())
 			}
 
+			// Skip if value (fee) is not greater than minimum bid
```
metachris, line 375:
```suggestion
			// Skip if value (fee) is less than minimum bid
```

metachris, line 375: would "minimum bid amount" imply that the value would already be accepted, rather than only values greater than it?

allboxes: You are right that does make more sense.  Updated to accept bids equal to the minimum.

cli/main.go
```diff
@@ -112,13 +115,25 @@ func Main() {
 		log.Fatal("Please specify a relay timeout greater than 0")
 	}
 
+
+	if *relayMinBidEth < 0.0 {
+		log.Fatal("Please specify a non-negative minimum bid")
+	}
+
+	if *relayMinBidEth > 100.0 {
```
erikleo, line 123: I don't see any reason to hardcode an arbitrary max minBid.  I think it should be up to the user to define any valid positive number.  Additionally mev boosts could be run a future forks where a high absolute number could be normal.

jumboshrimp100, line 123: I don't see any reason to hardcode an arbitrary max minBid. I think it should be up to the user to define any valid positive number. Additionally mev boost could be run on future forks where a high absolute number could be normal.

allboxes: The motivation was to catch user mistakes in the flag (e.g. entering the value in wei instead of eth).  I have no problem removing it or setting it much higher.

allboxes: Raised to 1,000,000 Ethers

allboxes:
@metachris do you need anything else from me on this?

cli/main.go
```diff
@@ -26,6 +28,7 @@ var (
 	defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
 	defaultRelayCheck         = os.Getenv("RELAY_STARTUP_CHECK") != ""
 	defaultGenesisForkVersion = getEnv("GENESIS_FORK_VERSION", "")
+	defaultRelayMinBidEth     = getEnvFloat64("RELAY_MIN_BID", 0.001)
```
metachris, line 31:
```suggestion
	defaultRelayMinBidEth     = getEnvFloat64("MIN_BID_ETH", 0.001)
```

cli/main.go
```diff
@@ -202,3 +220,31 @@ func getEnvInt(key string, defaultValue int) int {
...
+	if value, ok := os.LookupEnv(key); ok {
+		val, err := strconv.ParseFloat(value, 64)
+		if err == nil {
+			return val
+		}
+	}
+	return defaultValue
+}
+
+func floatEthTo256Wei(val float64) (*types.U256Str, error) {
```
metachris, line 234: maybe add some tests for this function

server/service.go
```diff
@@ -366,6 +369,11 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			// Remember which relays delivered which bids (multiple relays might deliver the top bid)
 			relays[BlockHashHex(blockHash)] = append(relays[BlockHashHex(blockHash)], relay)
 
+			// Skip if value (fee) is lower than the minimum bid
```
metachris, line 372: move this to line 365, above saving the relays and locking

metachris:
looks like a good feature to have. left a few comments, also please rebase 🙏

cli/main.go
allboxes: done

server/service.go
allboxes: done

cli/main.go
allboxes: Added a test but there are some potential rounding errors in the function, I don't think they impact the functionality but Ill update this after they are fixed.

server/service.go
```diff
@@ -358,7 +361,13 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 				return
 			}
 
+			if responsePayload.Data.Message.Value.Cmp(&m.relayMinBid) == -1 {
+				log.Debug("ignoring bid below min-bid value")
+				return
+			}
+
 			log.Debug("bid received")
+			// Skip if value (fee) is lower than the minimum bid
```
metachris, line 370: Move these two lines above to line 363

cli/main_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"testing"
+
+	"github.com/flashbots/go-boost-utils/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFloatEthTo256Wei(t *testing.T) {
+	// test with valid input
+	i := 0.000000000000012345
+	u256, err := floatEthTo256Wei(i)
```
metachris, line 13:
```suggestion
	wei, err := floatEthTo256Wei(i)
```

metachris: or `weiU256`

cli/main_test.go
```diff
@@ -0,0 +1,16 @@
...
+	require.Equal(t, types.IntToU256(12345), *u256)
+	require.NoError(t, err)
```
metachris, line 15: check error first, then the equal

allboxes:
Thanks for the comments, they should all be resolved in the above commit

jumboshrimp100:
@allboxes, I think this PR is step in the right direction given that many people are wary of running mev-boost especially since the mev-boost reward has the possibility of being less than the reward you would have earned from a locally produced block.  I have 2 comments (although this may be beyond the scope of the PR)

1) There are 2 types of min bid, absolute min bid and relative min bid.  I take it this PR is an absolute mid bid, but it would also be useful to have the option to specify a relative min bid which would be the min bid above the reward you would receive from a locally produced block

2) Is there a way of transmitting this information to the relay?  Under this implementation if I have a min bid of .1 eth will the relay realize this and only submit bids of .1 eth or more?  Will the bid immediately be rejected, giving the relay the opportunity to submit a higher bid?  Or will the low bid be silently ignored?  The last case seems sub-optimal if the relay is willing to bid higher but did not know of the min required bid of the validator.

cli/main.go
```diff
@@ -201,3 +219,46 @@ func getEnvInt(key string, defaultValue int) int {
...
+		}
+	}
+
+	weiInt := new(big.Int)
+	weiFloat.Int(weiInt)
+
+	err := weiU256.FromBig(weiInt)
+
+	return weiU256, err
+}
```
jtraglia, line 264: This method works just fine, but it's a little complicated in my opinion. Here's another way to do the same thing.

```suggestion
func floatEthTo256Wei(val float64) (*types.U256Str, error) {
	weiU256 := new(types.U256Str)
	ethFloat := new(big.Float)
	weiFloat := new(big.Float)
	weiFloatLessPrecise := new(big.Float)
	weiInt := new(big.Int)

	ethFloat.SetFloat64(val)
	weiFloat.Mul(ethFloat, big.NewFloat(1e18))
	weiFloatLessPrecise.SetString(weiFloat.String())
	weiFloatLessPrecise.Int(weiInt)

	err := weiU256.FromBig(weiInt)
	return weiU256, err
}
```

cli/main_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// test with large input
+	i = 987654.3
+	weiU256, err = floatEthTo256Wei(i)
+	require.NoError(t, err)
+
+	r := big.NewInt(9876543)
+	r.Mul(r, big.NewInt(100000000000000000)) // 10^17
+	refernceWeiU256 := new(types.U256Str)
+	err = refernceWeiU256.FromBig(r)
```
jtraglia, line 32:
```suggestion
	referenceWeiU256 := new(types.U256Str)
	err = referenceWeiU256.FromBig(r)
```

jtraglia (changes requested):
Overall, this looks pretty good. I like this feature. See my suggestions.

cli/main.go
allboxes: I had something similar to this originally but had issues with rounding errors.  This is much better, thanks!

jtraglia (dismissed):
LGTM. Thanks for fixing those things.

jtraglia (dismissed):
Thanks, I didn't see that typo.

cli/main.go
```diff
@@ -25,6 +27,7 @@ var (
 	defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
 	defaultRelayCheck         = os.Getenv("RELAY_STARTUP_CHECK") != ""
 	defaultGenesisForkVersion = getEnv("GENESIS_FORK_VERSION", "")
+	defaultRelayMinBidEth     = getEnvFloat64("MIN_BID_ETH", 0.001)
```
jtraglia, line 30: One more thing, the default minimum bid should be zero. It's up to the user (operator) to set this value based on their preferences. I'm just concerned with the side effects of this. For example, a default value of `0.001` would affect testnets (like [sepolia](https://builder-relay-sepolia.flashbots.net)) where the average mev reward is most likely to be below that threshold. I understand the desire for a non-zero default: low value blocks will be built using the local execution engine. But when CL clients start to compare the value to the block it built locally, I think that would serve as a better "default" minimum.

allboxes: Sure, that makes sense.

jtraglia (approved):
Looks great. Thanks!

metachris (approved):

PR #296, opened 2022-09-06
Add relay monitor URLs

avalonche:
## 📝 Summary

Start relay monitor support: https://github.com/ralexstokes/relay-monitor

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -67,8 +67,6 @@ type BoostService struct {
 
 	bidsLock sync.Mutex
 	bids     map[bidRespKey]bidResp // keeping track of bids, to log the originating relay on withholding
-
-	validatorRegistrationCh chan []types.SignedValidatorRegistration
 }
 
 // NewBoostService created a new BoostService
@@ -89,7 +87,6 @@ func NewBoostService(opts BoostServiceOpts) (*BoostService, error) {
 		log:                     opts.Log.WithField("module", "service"),
 		relayCheck:              opts.RelayCheck,
 		bids:                    make(map[bidRespKey]bidResp),
-		validatorRegistrationCh: make(chan []types.SignedValidatorRegistration),
 
 		builderSigningDomain: builderSigningDomain,
 		httpClientGetHeader: http.Client{
@@ -147,8 +144,6 @@ func (m *BoostService) StartHTTPServer() error {
 	}
 
 	go m.startBidCacheCleanupTask()
-	go m.startRelayMonitorValidatorTask()
-	defer close(m.validatorRegistrationCh)
 
 	m.srv = &http.Server{
 		Addr:    m.listenAddr,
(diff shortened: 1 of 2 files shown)


server/service.go
```diff
@@ -173,6 +182,28 @@ func (m *BoostService) startBidCacheCleanupTask() {
...
+	log := m.log.WithField("task", "relayMonitorTask")
+	log.Debug("registerValidator")
+	for {
+		payload := <-m.validatorRegistrationCh
+		log = log.WithField("numRegistrations", len(payload))
+		for _, relayMonitor := range m.relayMonitors {
+			go func(relayMonitor *url.URL) {
+				url := GetURI(relayMonitor, pathRegisterValidator)
+				log := m.log.WithField("url", url)
+				log.Info("requesting relay monitor")
```
metachris, line 195:
```suggestion
				log.Info("sending validator registration to relay monitor")
```

server/service.go
```diff
@@ -173,6 +182,28 @@ func (m *BoostService) startBidCacheCleanupTask() {
 	}
 }
 
+func (m *BoostService) startRelayMonitorValidatorTask() {
+	log := m.log.WithField("task", "relayMonitorTask")
+	log.Debug("registerValidator")
```
metachris, line 187:
```suggestion
	log.Debug("startRelayMonitorValidatorTask")
```

server/service.go
```diff
@@ -173,6 +182,28 @@ func (m *BoostService) startBidCacheCleanupTask() {
...
+			go func(relayMonitor *url.URL) {
+				url := GetURI(relayMonitor, pathRegisterValidator)
+				log := m.log.WithField("url", url)
+				log.Info("requesting relay monitor")
+				_, err := SendHTTPRequest(context.Background(), m.httpClientRegVal, http.MethodPost, url, UserAgent(""), payload, nil)
+				if err != nil {
+					log.WithError(err).Warn("error calling registerValidator on relay")
+					return
+				}
+				log.Info("success")
```
metachris, line 201:
```suggestion
				log.Info("sent validator registrations to relay monitor")
```

server/service.go
```diff
@@ -256,6 +287,8 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 		}(relay)
 	}
 
+	go func() { m.validatorRegistrationCh <- payload }()
```
metachris, line 290: i wonder if we shouldn't do a buffered channel instead, perhaps with a select to skip if full 🤔

metachris: i wonder if the channel is even needed, since it just kicks off a bunch of goroutines anyway.

this could be replaced with a function call, like 

```go
go m.sendValidatorRegistrationsToRelayMonitors(payload)
```

ralexstokes: you then move contention on the channel to contention for memory w/ a potentially unbounded number of goroutines...  I have made a note to ponder a similar issue with a use of channels in the relay-monitor... there is likely some common solution to handle the backpressure, it will just likely make the code look less nice ;)

metachris (approved):

JVillella:
Are there any docs on what a relay monitor is? More importantly, if it's not set MEV-Boost crashes with,

```
mev-boost[562791]: time="2022-09-09T21:55:12-04:00" level=warning msg="error calling registerValidator on relay monitor" error="Post \"/eth/v1/builder/validators\": unsupported protocol scheme \"\"" module=service url=/eth/v1/builder/validators
```

JVillella:
Great, thank you. And will give that doc a read.

PR #297, opened 2022-09-07
Send auction transcript to relay monitor

avalonche:
## 📝 Summary

Sends the signed builder bid and signed blinded beacon block to the relay monitor to verify the integrity of the auction. See https://hackmd.io/@ralexstokes/SynPJN_pq for the types of faults detected.

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -206,13 +206,13 @@ func (m *BoostService) sendValidatorRegistrationsToRelayMonitors(payload []types
 	}
 }
 
-func (m *BoostService) sendAuctionTranscriptToRelayMonitors(transcript AuctionTranscript) {
+func (m *BoostService) sendAuctionTranscriptToRelayMonitors(transcript *AuctionTranscript) {
 	log := m.log.WithField("method", "sendAuctionTranscriptToRelayMonitors")
 	for _, relayMonitor := range m.relayMonitors {
 		go func(relayMonitor *url.URL) {
 			url := GetURI(relayMonitor, pathAuctionTranscript)
 			log := log.WithField("url", url)
-			_, err := SendHTTPRequest(context.Background(), m.httpClientGetPayload, http.MethodPost, url, UserAgent(""), transcript, nil)
+			_, err := SendHTTPRequest(context.Background(), *http.DefaultClient, http.MethodPost, url, UserAgent(""), transcript, nil)
 			if err != nil {
 				log.WithError(err).Warn("error sending auction transcript to relay monitor")
 				return
@@ -490,6 +490,9 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 		log.Warn("bid found but no associated relays")
 	}
 
+	// send bid and signed block to relay monitor
+	go m.sendAuctionTranscriptToRelayMonitors(&AuctionTranscript{Bid: originalBid.response.Data, Acceptance: payload})
+
 	relays := originalBid.relays
 	if len(relays) == 0 {
 		log.Warn("originating relay not found, sending getPayload request to all relays")
@@ -566,8 +569,6 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 	// Wait for all requests to complete...
 	wg.Wait()
 
-	// send bid and signed block to relay monitor
-	m.sendAuctionTranscriptToRelayMonitors(AuctionTranscript{Bid: originalBid.response.Data, Acceptance: payload})
 	// If no payload has been received from relay, log loudly about withholding!
 	if result.Data == nil || result.Data.BlockHash == nilHash {
 		originRelays := RelayEntriesToStrings(originalBid.relays)


server/backend.go
```diff
@@ -1,9 +1,12 @@
...
-// Router paths
 var (
+	// Router paths
 	pathStatus            = "/eth/v1/builder/status"
 	pathRegisterValidator = "/eth/v1/builder/validators"
 	pathGetHeader         = "/eth/v1/builder/header/{slot:[0-9]+}/{parent_hash:0x[a-fA-F0-9]+}/{pubkey:0x[a-fA-F0-9]+}"
 	pathGetPayload        = "/eth/v1/builder/blinded_blocks"
+
+	// Relay Monitor paths
+	pathAuctionTranscript = "/eth/v1/relay-monitor/transcript"
```
metachris, line 11: TBD whether the `/eth/` namespace should be reused (i think perhaps not). cc/ @ralexstokes

ralexstokes, line 11:
```suggestion
	pathAuctionTranscript = "/api/v1/relay-monitor/transcript"
```

metachris: how about `/relay-monitor/v1/transcript`?

ralexstokes: im open to this -- the idea was to namespace the API of the relay monitor from any other HTTP endpoints that software may host, e.g. landing page

but I'm not super attached to the actual symbol

metachris: I just thought it'd be cleaner to keep /eth/ for eth specs, like the builder API. Not a strong opinion.

Cc @jtraglia @lightclient @terencechain ?

server/service.go
```diff
@@ -512,6 +534,8 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 	// Wait for all requests to complete...
 	wg.Wait()
 
+	// send bid and signed block to relay monitor
+	go m.sendAuctionTranscriptToRelayMonitors(AuctionTranscript{Bid: originalBid.response.Data, Acceptance: payload})
```
Ruteri, line 538: Since `sendAuctionTranscriptToRelayMonitors` is just spawning goroutines, I'd consider just calling it synchronously

server/backend.go
ralexstokes: the current thing is `api` and not `eth` -- this doesn't touch any of the "canonical" stuff at all and was just an arbitrary choice I made to namespace these endpoints

and I'm saying it can be `relay-monitor` if you want -- I don't think it is super critical

jtraglia: It's not a strong preference, but I prefer `/relay-monitor/v1/transcript` over the other options. Though, I suggest replacing the dash with an underscore (`/relay_monitor/v1/transcript`) as dashes aren't used Ethereum APIs or shortening it to `/monitor/v1/transcript` (this looks cleaner to me).

metachris: I like it. +1 for `/monitor/`

server/service.go
metachris: the calls to the monitor are spawned in goroutines, so slow monitors wouldn't affect this call (mateusz mentioned only doing this call sync, but still spawning the individual calls as goroutines)

server/service_test.go
```diff
@@ -478,7 +478,7 @@ func TestGetPayload(t *testing.T) {
 	})
 
 	t.Run("Bad response from relays", func(t *testing.T) {
-		backend := newTestBackend(t, 2, time.Second)
+		backend := newTestBackend(t, 2, 2*time.Second)
```
metachris, line 481: this seems not needed anymore! please revert

server/backend.go
ralexstokes: https://github.com/ralexstokes/relay-monitor/pull/33

server/service.go
```diff
@@ -200,6 +206,22 @@ func (m *BoostService) sendValidatorRegistrationsToRelayMonitors(payload []types
 	}
 }
 
+func (m *BoostService) sendAuctionTranscriptToRelayMonitors(transcript AuctionTranscript) {
+	log := m.log.WithField("method", "sendAuctionTranscriptToRelayMonitors")
+	for _, relayMonitor := range m.relayMonitors {
+		go func(relayMonitor *url.URL) {
+			url := GetURI(relayMonitor, pathAuctionTranscript)
+			log := log.WithField("url", url)
+			_, err := SendHTTPRequest(context.Background(), m.httpClientGetPayload, http.MethodPost, url, UserAgent(""), transcript, nil)
```
ralexstokes, line 215: do we want to use this particular `http.Client` here? I see this is logically under the `getPayload` path but im not sure that is what we want long term... could be fine for now and just something to keep in mind if it becomes a hotspot

server/service.go
```diff
@@ -200,6 +206,22 @@ func (m *BoostService) sendValidatorRegistrationsToRelayMonitors(payload []types
 	}
 }
 
+func (m *BoostService) sendAuctionTranscriptToRelayMonitors(transcript AuctionTranscript) {
```
ralexstokes, line 209: nitpick but we could pass in a pointer here
```suggestion
func (m *BoostService) sendAuctionTranscriptToRelayMonitors(transcript *AuctionTranscript) {
```

I tend to do so unless it is clear the input will *always* be small enough

server/service.go
```diff
@@ -544,6 +566,8 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
 	// Wait for all requests to complete...
 	wg.Wait()
 
+	// send bid and signed block to relay monitor
+	m.sendAuctionTranscriptToRelayMonitors(AuctionTranscript{Bid: originalBid.response.Data, Acceptance: payload})
```
ralexstokes, line 570: if we change the type signature above:

```suggestion
	m.sendAuctionTranscriptToRelayMonitors(&AuctionTranscript{Bid: originalBid.response.Data, Acceptance: payload})
```

ralexstokes, line 570: is there a way to return early before this is fired?

`mev-boost` should post this data to any relay monitors if it gets the data called `payload` here, regardless of any relay behavior -- and there is a large chunk of code to manage that leg of this journey in between the start of this function and making the call to the relay monitor

ralexstokes (commented):
left a few comments, overall looks great!

server/service.go
metachris: yeah. this function does just spawn off a few more goroutines, but that could change in the future, so it'd be better to keep the `go m.sendAuctionTranscriptToRelayMonitors` here.

server/service.go
metachris: +1

ralexstokes (approved):
nice updates!

PR #294, opened 2022-09-05
Add prometheus metrics endpoint

0xSte:
## 📝 Summary

An initial implementation of some prometheus metrics for MEV Boost

## ⛱ Motivation and Context

I want to be able to monitor my mev-boost container with prometheus to get better operational insight

## 📚 References

https://github.com/flashbots/mev-boost/issues/253
---

## ✅ I have run these commands

* [✔️] `go mod tidy`
* [✔️] `make test`
* [✔️] `make lint `
* [❔] `make run-mergemock-integration`

Getting an error back from `make run-mergemock-integration`, but not sure if it's related to my change
```
ERROR  [2022-09-05T19:09:20+01:00] Unable to retrieve proposal payload           error="builder REST API returned non-200 status code: 204" slot="5"
make: *** [run-mergemock-integration] Error 1```

PR #305, opened 2022-09-10
WIP: (Feedback wanted) Log relay and value of received bids

pietjepuk2:
## 📝 Summary

Just a quick logging of bid value. This way people can check that, when the have proposed a blinded block, the value they got out of it matches the value communicated by the relay. We're already logging information about the bid being e.g. zero (with a warning), or that we received a valid new bid (with a debug message), so it makes sense to include logging on the actual bid as well. 

#294 is probably a more "complete" solution (with timings and everything), but this was such a simple one-liner that I couldn't resist.

## TODO:
- [x] Determine log level (info, or debug?)
- [ ] Elaborate the _why_ in commit message 

## 📚 References

---

* [ ] `make lint`
* [ ] `make test-race`
* [ ] `go mod tidy`
"

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -408,6 +408,9 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			// Compare the bid with already known top bid (if any)
 			if result.response.Data != nil {
 				valueDiff := responsePayload.Data.Message.Value.Cmp(&result.response.Data.Message.Value)
+
+				log.Infof("Relay %s offered bid with value %s", relay.String(), responsePayload.Data.Message.Value.String())
+
 				if valueDiff == -1 { // current bid is less profitable than already known one
 					return
 				} else if valueDiff == 0 { // current bid is equally profitable as already known one. Use hash as tiebreaker
--- a/server/service.go
+++ b/server/service.go
@@ -421,7 +421,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			}
 
 			// Use this relay's response as mev-boost response because it's most profitable
-			log.Debug("selecting bid")
+			log.Debug("new best bid")
 			result.response = *responsePayload
 			result.blockHash = blockHash
 			result.t = time.Now()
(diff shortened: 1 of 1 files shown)


server/service.go
```diff
@@ -408,6 +408,9 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			// Compare the bid with already known top bid (if any)
 			if result.response.Data != nil {
 				valueDiff := responsePayload.Data.Message.Value.Cmp(&result.response.Data.Message.Value)
+
+				log.Infof("Relay %s offered bid with value %s", relay.String(), responsePayload.Data.Message.Value.String())
```
metachris, line 412: relay and value are already added as fields to `log` at this point, and to reduce verbosity I'd change this to `log.Debug`. So the function would look like `log.Debug("bid received")

metachris: you probably also want to log this above the `if` in 409, because currently it would only log if it's not the first received bid

pietjepuk2: > relay and value are already added as fields to `log` at this point, and to reduce verbosity I'd change this to `log.Debug`. So the function would look like `log.Debug("bid received")`

Ah, so that's what the `log.withField` stuff is for. Noted, and indeed will just change it to a simple "bid received" then :)

 > you probably also want to log this above the if in 409, because currently it would only log if it's not the first received bid
And that's why I shouldn't do late-night programming, well-caught!

I do wonder about the `log.Debug("received a good bid")` at the end now though. Something along the lines of "using bid as response" or "selecting bid" would be more descriptive and signify the difference with the new log message. I now picked the latter, because it's the shortest.

server/service.go
```diff
@@ -419,7 +421,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			}
 
 			// Use this relay's response as mev-boost response because it's most profitable
-			log.Debug("received a good bid")
+			log.Debug("selecting bid")
```
metachris, line 424: at this point it's not selected. it's just better than a previous one. i'd leave that unchanged, or reword somehow differently.

pietjepuk2: Oh, I was wording that based on the comment above that line. Does that mean that the comment is wrong, or that "selecting" means something other than "this is what we will respond with"?

The wording "than a previous one" implies there was a previous one, which is not necessarily the case right?

metachris:
thanks 👍

PR #329, opened 2022-09-18
check for new releases on github, and log a warn if update is possible

metachris:
## 📝 Summary

This is a proposal and request for comments and input.

* check for new mev-boost releases on Github regularly, and log (info) if update is possible
* can be disabled with setting env var `DISABLE_NEW_RELEASE_CHECK=1` 
* default check time: at startup and every 2 hours
* timeout can be configured with `NEW_RELEASE_CHECK_INTERVAL_H`

**Question:** should this be opt-in or opt-out?

## ⛱ Motivation and Context

mev-boost users should get notified if new releases are available

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


Ruteri (approved):

jtraglia:
I like the idea but I'm not sure how useful it would be in practice. The type of person that this caters to probably isn't the type of person that regularly checks their mev-boost logs, only when something isn't working properly.

metachris:
closing this for now. may be worth reconsidering with an opt-in flag

PR #334, opened 2022-09-20
Misc Update README.md

nmoreau:
## 📝 Summary

It was unclear from that doc that we needed to update the configuration of both the beacon and validator. I made it clearer.

## ⛱ Motivation and Context
I was not able to find clear instruction at first glance when setting up MEV Boost on my validator. I guess others may also be tricked.

## 📚 References

---

* [X ] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ With MEV-Boost, validators can access blocks from a marketplace of builders. Bui
 ## How does MEV-Boost work?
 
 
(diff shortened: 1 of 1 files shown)


metachris:
thanks. merged into a temporary branch to add and fix a little, will get into a new PR soon 🙏

Issue #323, opened 2022-09-16
feature-request: retry getPayload on error (until configured timeout)


metachris:
getPayload should retry the call on errors, as long as it's below the configured timeout

metachris:
done, available in mev-boost v1.5.0-alpha1

PR #340, opened 2022-09-21
Add retries for get payload

avalonche:
## 📝 Summary

Resolves issue https://github.com/flashbots/mev-boost/issues/323 allowing getPayload to retry the call on errors, as long as it's below the configured timeout

## ⛱ Motivation and Context

Improves reliability of the getPayload call

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -62,6 +62,7 @@ type BoostServiceOpts struct {
 	RequestTimeoutGetHeader  time.Duration
 	RequestTimeoutGetPayload time.Duration
 	RequestTimeoutRegVal     time.Duration
+	RequestMaxRetries        int
 }
 
 // BoostService - the mev-boost service
@@ -78,6 +79,7 @@ type BoostService struct {
 	httpClientGetHeader  http.Client
 	httpClientGetPayload http.Client
 	httpClientRegVal     http.Client
+	requestMaxRetries    int
 
 	bidsLock sync.Mutex
 	bids     map[bidRespKey]bidResp // keeping track of bids, to log the originating relay on withholding
@@ -116,6 +118,7 @@ func NewBoostService(opts BoostServiceOpts) (*BoostService, error) {
 			Timeout:       opts.RequestTimeoutRegVal,
 			CheckRedirect: httpClientDisallowRedirects,
 		},
+		requestMaxRetries: opts.RequestMaxRetries,
 	}, nil
 }
 
@@ -501,12 +504,9 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
 
 			log := log.WithField("url", url)
 			log.Debug("calling getPayload")
-			errorHandler := func(err error) {
-				log.WithError(err).Debug("error making request to relay, retrying")
-			}
 
 			responsePayload := new(types.GetPayloadResponse)
-			_, err := SendHTTPRequestWithRetries(requestCtx, m.httpClientGetPayload, http.MethodPost, url, ua, payload, responsePayload, errorHandler)
+			_, err := SendHTTPRequestWithRetries(requestCtx, m.httpClientGetPayload, http.MethodPost, url, ua, payload, responsePayload, m.requestMaxRetries, log)
 			if err != nil {
 				if errors.Is(requestCtx.Err(), context.Canceled) {
 					log.Info("request was cancelled") // this is expected, if payload has already been received by another relay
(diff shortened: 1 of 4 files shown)


server/utils.go
```diff
@@ -77,6 +80,38 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
...
+	if client.Timeout == 0 {
+		return SendHTTPRequest(ctx, client, method, url, userAgent, payload, dst)
+	}
+
+	// Create a context with a timeout as configured in the http client
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), client.Timeout)
+	defer cancel()
+
+	attempts := 0
+	retryWaitTime := 1 * time.Second
```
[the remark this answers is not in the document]
metachris: i think it should not wait, or maybe 100ms. there's not a lot of time to loose. if wait time then yes, configurable as env var at least

server/utils.go
```diff
@@ -77,6 +80,38 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
...
+	for {
+		select {
+		case <-ctxTimeout.Done():
```
metachris, line 98: fyi, this will not interrupt a running request because it's blocking in the `default:` part, the current code works just like a for loop without select.

metachris, line 98: this will not interrupt the request because it's already blocking. so instead of using a `select` here it might be easier to understand to write this function as a simple for loop without the select.

server/utils.go
```diff
@@ -77,6 +80,38 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
 	return resp.StatusCode, nil
 }
 
+// SendHTTPRequestWithRetries - prepare and send HTTP request, retrying the request if within the client timeout
+func SendHTTPRequestWithRetries(ctx context.Context, client http.Client, method, url string, userAgent UserAgent, payload any, dst any, handleError ErrorHandlerFunc) (code int, err error) {
+	// default no retry if no timeout set
+	if client.Timeout == 0 {
```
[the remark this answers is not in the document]
metachris: yes, perhaps.

server/service.go
```diff
@@ -498,11 +498,15 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
...
+
 			log := log.WithField("url", url)
 			log.Debug("calling getPayload")
+			errorHandler := func(err error) {
+				log.WithError(err).Debug("error making request to relay, retrying")
+			}
 
 			responsePayload := new(types.GetPayloadResponse)
-			_, err := SendHTTPRequest(requestCtx, m.httpClientGetPayload, http.MethodPost, url, ua, payload, responsePayload)
+			_, err := SendHTTPRequestWithRetries(requestCtx, m.httpClientGetPayload, http.MethodPost, url, ua, payload, responsePayload, errorHandler)
```
metachris, line 509: should we also do this for `processCapellaPayload`?

server/service.go
```diff
@@ -498,11 +498,15 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
 		go func(relay RelayEntry) {
 			defer wg.Done()
 			url := relay.GetURI(pathGetPayload)
+
 			log := log.WithField("url", url)
 			log.Debug("calling getPayload")
+			errorHandler := func(err error) {
+				log.WithError(err).Debug("error making request to relay, retrying")
```
metachris, line 505: this should probably be an error log, at least warn, so users see it
```suggestion
				log.WithError(err).Warn("error making request to relay, retrying")
```

server/utils.go
```diff
@@ -91,6 +100,39 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
 	return resp.StatusCode, nil
 }
 
+// SendHTTPRequestWithRetries - prepare and send HTTP request, retrying the request if within the client timeout
+func SendHTTPRequestWithRetries(ctx context.Context, client http.Client, method, url string, userAgent UserAgent, payload, dst any, handleError ErrorHandlerFunc) (code int, err error) {
```
metachris, line 104: it may be simpler to pass in the logger instance, and log from inside, rather than pass in an error handler?

server/utils.go
```diff
@@ -23,10 +23,16 @@ import (
 	"github.com/flashbots/mev-boost/config"
 )
 
+const (
+	// DefaultMaxRetries is the default number of retries for HTTP requests
+	DefaultMaxRetries = 5
```
jtraglia, line 28: Do you want this to be a configurable option? To me, default implies there's a way to change it.

jtraglia (approved):
LGTM 👍

jtraglia (approved):
Still LGTM 👍

0xpanoramix (approved):

PR #342, opened 2022-09-22
getPayload only to origin relays

metachris:
## 📝 Summary

Send `getPayload` requests only to the relays that sent the bid

## ⛱ Motivation and Context

Currently getPayload is sent to all relays, which will result in a bunch of errors in the logs because most relays won't know about the requested payload.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -286,13 +286,11 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
-	var mu sync.Mutex
-	relays := make(map[string][]string) // relays per blockHash
-	result := bidResp{}
-
-	ua := UserAgent(req.Header.Get("User-Agent"))
+	result := bidResp{}                           // the final response, containing the highest bid (if any)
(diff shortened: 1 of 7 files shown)


server/service.go
```diff
@@ -287,7 +287,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	}
 
 	var mu sync.Mutex
-	relays := make(map[string][]string) // relays per blockHash
+	relays := make(map[string][]RelayEntry) // relays per blockHash
```
Ruteri, line 290: This makes me happy. Should the comment go?
I could sign off to `relays := make(map[BlockHash][]RelayEntry)`

Ruteri (approved):

server/service.go
metachris: done

Ruteri (approved):

metachris:
fwiw, confirmed working on sepolia, thanks Kiln 🙏

PR #348, opened 2022-09-23
Filter out duplicate relays & monitors

jtraglia:
## 📝 Summary

Noticed that there were no duplication checks. You could specify the same relay or monitor multiple times.

## ⛱ Motivation and Context

May prevent unnecessary work by the MEV client and relays.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -189,7 +189,7 @@ func unique(stringSlice []string) []string {
 	filtered := make([]string, 0, len(stringSlice))
 	stringMap := make(map[string]bool)
 	for _, entry := range stringSlice {
-		if _, s := stringMap[entry]; !s {
+		if !stringMap[entry] {
 			filtered = append(filtered, entry)
 			stringMap[entry] = true
 		}
--- a/cli/main.go
+++ b/cli/main.go
@@ -77,7 +77,6 @@ func Main() {
 		log.Logger.SetFormatter(&logrus.TextFormatter{
 			FullTimestamp: true,
 		})
-
 	}
 
 	if *logDebug {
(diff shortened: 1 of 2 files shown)


jtraglia:
Also, without this PR, `mev-boost` will return a not-so-useful error if the `-relays` flag isn't set:

```
$ make build && ./mev-boost -mainnet  
go build -trimpath -ldflags "-s -X 'github.com/flashbots/mev-boost/config.Version=v1.3.2-5-gece7a08'" -v -o mev-boost .
github.com/flashbots/mev-boost
INFO[2022-09-23T13:59:38-05:00] mev-boost v1.3.2-5-gece7a08                  
INFO[2022-09-23T13:59:38-05:00] using genesis fork version: 0x00000000       
FATA[2022-09-23T13:59:38-05:00] Invalid relay URL                             error="missing relay public key" relayURL=
```

This is because `relayURLs` is an empty string (`""`) when not specified. That's its default value.

With this PR, it will print the "no relays specified" error like we want.

metachris:
a test or two would be very nice for this

jtraglia:
Yeah tests are a good idea. Added some. I would have done some relay monitor tests but it very repetitive since it's essentially the same thing. Also, it's not really possible to test invalid relay cases because `parseRelayURLs` will immediately exit without returning an error if something is wrong, like a missing pubkey.

cli/main.go
```diff
@@ -173,9 +172,33 @@ func getEnvInt(key string, defaultValue int) int {
 	return defaultValue
 }
 
-func parseRelayURLs(relayURLs string) []server.RelayEntry {
-	ret := []server.RelayEntry{}
-	for _, entry := range strings.Split(relayURLs, ",") {
+// Deletes whitespace from beginning and end, removes empty strings
+func trim(stringSlice []string) []string {
+	filtered := make([]string, 0, len(stringSlice))
```
metachris, line 177: since the final size is not known, don't initialize the slice with a size?
```suggestion
	filtered := []string{}
```

cli/main.go
```diff
@@ -173,9 +172,33 @@ func getEnvInt(key string, defaultValue int) int {
...
+		if trimmed != "" {
+			filtered = append(filtered, trimmed)
+		}
+	}
+	return filtered
+}
+
+// Returns a slice of unique items
+func unique(stringSlice []string) []string {
+	filtered := make([]string, 0, len(stringSlice))
```
metachris, line 189:
```suggestion
	filtered := []string{}
```

the final size is not known, don't initialize it

cli/main.go
```diff
@@ -173,9 +172,33 @@ func getEnvInt(key string, defaultValue int) int {
...
+	stringMap := make(map[string]bool)
+	for _, entry := range stringSlice {
+		if _, s := stringMap[entry]; !s {
```
metachris, line 192:
```suggestion
		if !stringMap[entry] {
```
could be simplified like this

cli/main.go
jtraglia: So that third argument is the capacity, not size. This is an optimization.

```go
func make([]T, len, cap) []T
```

Since we know the maximum number of items there can be (the length of the input slice), I think this is reasonable. Pre-allocations can speed things up quite a bit, not that this is a time sensitive operation.

Here's some random article that explains it pretty well:

* https://medium.com/swlh/bad-go-not-sizing-slices-aed1b01cff83

cli/main.go
jtraglia: See my other comment.

jtraglia:
If #351 gets merged, this PR is no longer necessary.

PR #353, opened 2022-09-25
style: remove old relays from readme

bertmiller:
Removes kiln and ropsten from the list of testnets that are supported for testing

No other changes - in particular code changes - are made


metachris:
Thanks

PR #358, opened 2022-09-26
collapsible markup bug - Systemd configuration

kailinr:
The ## Systemd configuration <details> tag is not closing properly. Rest of the content is nested under the dropdown. Suggest removing it or fixing the collapsible tagging.

## 📚 References

---

* [ ] `make lint`
* [ ] `make test-race`
* [ ] `go mod tidy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,8 +145,7 @@ docker run flashbots/mev-boost -help
 
 You can run MEV-Boost with a systemd config like this:
 
-<details>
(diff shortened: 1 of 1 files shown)


jtraglia:
#360 also fixes this problem, but keeps the collapsable details section.

metachris:
thanks, but we want it collapsable

PR #360, opened 2022-09-27
Fix systemd configuration section in readme

jtraglia:
## 📝 Summary

This section isn't showing up correctly and is preventing the end of the readme from loading:

image

* Add a space between `<summary>` and code block.
* Use the `<code>` tag instead of backticks because those don't work in HTML tags.

## ⛱ Motivation and Context

Need to fix this so the whole readme is visible.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,8 @@ docker run flashbots/mev-boost -help
 You can run MEV-Boost with a systemd config like this:
 
 <details>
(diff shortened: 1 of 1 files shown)


metachris:
thanks!

PR #362, opened 2022-09-28
log version always, fix loglevel config

metachris:
## 📝 Summary

* always log the mev-boost version (simplifies debugging)
* fix custom loglevel (bug introduced in a previous commit, but was not released
* opt-out flag: `-log-no-version` / env var `DISABLE_LOG_VERSION`

```
INFO[2022-09-29T21:42:42+02:00] starting mev-boost                            version=v1.3.3-dev
DEBU[2022-09-29T21:42:42+02:00] debug logging enabled                         version=v1.3.3-dev
INFO[2022-09-29T21:42:42+02:00] using genesis fork version: 0x00000000        version=v1.3.3-dev
INFO[2022-09-29T21:42:42+02:00] using 6 relays                                version=v1.3.3-dev
```

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -26,11 +26,15 @@ var (
 	defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
 	defaultRelayCheck         = os.Getenv("RELAY_STARTUP_CHECK") != ""
 	defaultGenesisForkVersion = getEnv("GENESIS_FORK_VERSION", "")
+	defaultDisableLogVersion  = os.Getenv("DISABLE_LOG_VERSION") == "1" // disables adding the version to every log entry
 
 	// mev-boost relay request timeouts (see also https://github.com/flashbots/mev-boost/issues/287)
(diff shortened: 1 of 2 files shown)


jtraglia:
I agree that printing the version on every line would help with debugging when given random log excerpts from users, but I'm not a fan of it being appended to the end of line. I think a better place for the version string would be between the timestamp and the message. Something like:

```
INFO[2022-09-28T09:10:06-05:00][v1.3.2] using genesis fork version: 0x00000000
```

To do this, we would need to define a new formatter. See:
* https://github.com/Sirupsen/logrus#formatters

Thoughts?

jtraglia:
Also, with this PR, the log that prints the first during startup is unnecessary and can be removed:

```
INFO[2022-09-28T09:10:06-05:00] mev-boost v1.3.2-17-ge2411c2                  version=v1.3.2-17-ge2411c2
```

jtraglia (dismissed):

metachris:
Thanks for your comments, Justin. I agree it's a tradeoff. But it will really help with debugging so leaning towards accepting the ux. Maybe we should just add an opt-out flag!

jtraglia (approved):
LGTM!

PR #365, opened 2022-10-01
accept gzip encoding

dmarzzz:
## 📝 Summary

From a discussion with @zeroecco on how to lower bandwidth on the blocknative relay we came to the idea that we could utilize well known compression like gzip, but only if requests coming from mev-boost specified that.

Minimum transmittable unit is 1500 Bytes total w/ 1250 Bytes realistically usable and a block is around ~12KB, gzip will compress ~60% to ~6Kb which would be a packet reduction from 10 to 5 for the total packet train. This also reduces time to last packet and bandwidth, which could be 1 - 5ms based on response time at this size.

If a system doesn't have this enabled the content-encoding will simply be ignored.
On top of that, this improvement will really only be visible for GetPayload response.

To benefit from this in your relay or other components make sure you configure your reverse proxy [HA Proxy, NGINX, etc] to support gzip!

## ⛱ Motivation and Context

I want to go fast

## 📚 References

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

https://www.rfc-editor.org/rfc/rfc2616#section-3.5

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -26,11 +26,15 @@ var (
 	defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
 	defaultRelayCheck         = os.Getenv("RELAY_STARTUP_CHECK") != ""
 	defaultGenesisForkVersion = getEnv("GENESIS_FORK_VERSION", "")
+	defaultDisableLogVersion  = os.Getenv("DISABLE_LOG_VERSION") == "1" // disables adding the version to every log entry
 
 	// mev-boost relay request timeouts (see also https://github.com/flashbots/mev-boost/issues/287)
 	defaultTimeoutMsGetHeader         = getEnvInt("RELAY_TIMEOUT_MS_GETHEADER", 950)   // timeout for getHeader requests
 	defaultTimeoutMsGetPayload        = getEnvInt("RELAY_TIMEOUT_MS_GETPAYLOAD", 4000) // timeout for getPayload requests
 	defaultTimeoutMsRegisterValidator = getEnvInt("RELAY_TIMEOUT_MS_REGVAL", 3000)     // timeout for registerValidator requests
+
+	relays        relayList
+	relayMonitors relayMonitorList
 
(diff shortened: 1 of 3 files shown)


metachris:
Could you add a little bit of context please?

zeroecco:
Internally our testing did show a time to last packet reduction in deliveries of blocks of roughly 1-5ms depending on the block size and the resources of the validator. 

Additionally this change can lead to more time for validators to propose on chain and thereby potentially lead fewer missing proposals though more efficient packet transport over the wider internet.

dmarzzz:
added some more context @metachris , didn't realize drafts showed up in public repos whoops!

metachris:
Thanks. I suppose you'd still need to add gzip decoding into the request processing? Like outlined here: https://stackoverflow.com/a/13131806

dmarzzz:
Awesome good fix! I added a note to the PR as well that for relays and others to enable this make sure you configure your reverse proxy [HA Proxy, NGINX, etc] to support gzip!

dmarzzz:
The test that lint is tripping up on works for me locally 🤔

metachris:
I think the test is because of an older staticcheck version, will push a PR

metachris:
https://github.com/flashbots/mev-boost/pull/374

metachris:
@dmarzzz could you add a test where the mockserver returns a gzipped response?

jtraglia (approved):
LGTM. Nice addition. Thanks!

metachris:
Added a test for the gzip response: https://github.com/dmarzzz/mev-boost/pull/1 (needs to be merged by @dmarzzz into his repo to update this PR)

dmarzzz:
@metachris thanks appreciate it, wasn't able to get to it till this weekend. When I merged it auto dismissed @jtraglia review

jtraglia (approved):
No worries. Thanks for the extra test.

PR #366, opened 2022-10-04
getHeader: print bid value in eth instead of wei

metachris:
## 📝 Summary

Request for discussion

This PR prints the getHeader bid values in ETH instead of wei (which is a popular community request)

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/types.go
+++ b/cli/types.go
@@ -1,10 +1,11 @@
 package cli
 
 import (
-	"github.com/flashbots/mev-boost/server"
-	"strings"
-	"net/url"
 	"errors"
+	"net/url"
+	"strings"
+
+	"github.com/flashbots/mev-boost/server"
 )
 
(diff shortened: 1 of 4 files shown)


jtraglia (approved):
Nice. Looks good to me. I'm a fan of this change.

I fuzzed the function for a little bit too. Used this method:

```go
func FuzzWeiBigIntToEthBigFloat(f *testing.F) {
	f.Fuzz(func(t *testing.T, data []byte) {
		tp, err := go_fuzz_utils.NewTypeProvider(data)
		if err != nil {
			return
		}
		var value types.U256Str
		err = tp.Fill(&value)
		if err != nil {
			return
		}
		t.Logf("Value: %s", value.BigInt())
		weiBigIntToEthBigFloat(value.BigInt())
	})
}
```

No findings. You're welcome to add that to the tests if you want.

PR #367, opened 2022-10-06
Enable more golangci checks & remove revive

jtraglia:
## 📝 Summary

* Essentially, copy `.golangci.yml` file from go-boost-utils.
  * By default, everything is enabled. Then we disable checks we don't like.
* Remove revive as it's run by golangci anyway.
* Copy `make lint` rule from go-boost-utils/mev-boost-relay.
  * This calls `gofumpt` with extra checks.
  * And gofmt with simplify.
* Fix findings from a few different linters:
  * Use go1.13 style errors.
  * Replace `fmt.Info` calls with `log.Info` calls.
  * Print version after the `SetFormatter` call.
  * Fix bug in `relayMonitorList.String()` that `makezero` identified.
  * Call `t.Helper()` at the beginning of helper functions.
  * Replace `"POST"` with `http.MethodPost` in `http.NewRequest` calls.

## ⛱ Motivation and Context

Just want to keep things consistent between the three projects.

I may review the disabled checks again later and try to fix some more stuff.

Fixes #306.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,9 +28,6 @@ jobs:
           git diff-index HEAD
           git diff-index --quiet HEAD
 
-      - name: Install revive linter
-        run: go install github.com/mgechev/revive@v1.1.3
-
       - name: Install gofumpt
         run: go install mvdan.cc/gofumpt@v0.3.1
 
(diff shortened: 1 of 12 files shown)


metachris:
Great QoL improvements, thanks 🙏

metachris (approved):

PR #376, opened 2022-10-19
readme: clarify that usage is about main branch

metachris:
## 📝 Summary

* Make it clear in the README that the usage instructions reflect the `main` branch, not the latest release
* Add generating usage to release instructions

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


jtraglia (approved):
LGTM 👍 though, personally, I don't think the help output should be in the readme.

metachris:
Maybe we should remove the help output from the readme and only have it in the releases. But i've heard from the community that this seems useful, although confusing if it doesn't match the latest release. But this still applies to the other usage instructions as well 🤔 

If you feel strongly about removing the help output from the readme, them please open a PR, and add a link to the releases page in the readme with the note that cli usage instructions are found there. i think i agree that this would be the better choice 🙏

PR #379, opened 2022-10-20
Minor release fixes for v1.4.0

metachris:
## 📝 Summary

* Github CI: on pushing tag with `-` (i.e. `-rc1` suffix), don't override Docker images for major.minor and major builds, but only build tags for sha and specific version
* `mev-boost -version` should print the version without the logging, as is customary
* Updated RELEASE.md
* Less verbose codecov report

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,11 +27,11 @@ jobs:
         with:
           images: flashbots/mev-boost
           tags: |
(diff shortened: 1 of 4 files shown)


jtraglia (approved):
LGTM 👍

PR #381, opened 2022-10-21
use mainnet by default

metachris:
## 📝 Summary

Use `-mainnet` by default

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


terencechain (approved):
nice!

metachris:
added a fix, otherwise it wouldn't allow any other network :)

terencechain (approved):

PR #382, opened 2022-10-21
Gzip fix

metachris:
## 📝 Summary

Fix bugs in requesting gzip encoded response (#365. cc/ @dmarzzz)

* Always set the accept-encoding header, not only if there's a request payload
* Manually add the gzip reader

Docker image: v1.4.0-rc3 

Tested successfully in Sepolia by Kiln.fi

---

## ✅ I have run these commands

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

PR #383, opened 2022-10-21
remove manual gzipping of request, as it's done automatically!

metachris:
## 📝 Summary

Fixes #365 and replaces #382

http.Request automatically adds `Accept-Encoding: gzip` header and transparently decompresses the response. You can see this working in the test in this PR. Also see https://pkg.go.dev/net/http#Response.Uncompressed

But when the header is set manually, then it won't decode the response automatically, which is what we're seeing in #382 

In fact, gzip was already support since the very beginning of mev-boost. :)

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -45,7 +45,6 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
 
 		// Set headers
 		req.Header.Add("Content-Type", "application/json")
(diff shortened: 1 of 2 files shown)


metachris:
Merging this quickly because `getPayload` on the main branch is currently broken

PR #389, opened 2022-10-28
Add bug bounty section to readme

jtraglia:
## 📝 Summary

During the roundtable at Bogota, someone mentioned that the bug bounty program wasn't on the README when it should have been. This PR moves that section from SECURITY to README. There should be more details about the bug bounty program in that security file though 🔒

Also, I think the details this bug bounty program are pretty vague. What does a "shared pool of $50k" mean exactly? I understand that's the total allocation for the program, but what are the actual expected bounties? If someone finds a critical vulnerability, will they get all of that $50k? A table of maximum bounties would be nice, like:

image

## ⛱ Motivation and Context

More exposure for the bug bounty 👍 

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ See also:
 - [Maintainers](#maintainers)
 - [Contributing](#contributing)
 - [Security](#security)
+  - [Bug Bounty](#bug-bounty)
   - [Audits](#audits)
 - [License](#license)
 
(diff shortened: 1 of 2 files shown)


jtraglia:
For example, something like:


| Severity |     Maximum | Example                                        |
|----------|------------:|------------------------------------------------|
| Low      |    $200 USD | A bug that rejects valid bids.                 |
| Medium   |  $1,000 USD | Can crash mev-boost via builder message.       |
| High     |  $5,000 USD | Can cause a connected validator to go offline. |
| Critical | $25,000 USD | Can remotely access arbitrary files on host.   |

These maximums are EF bounty program's maximums divided by 10. Examples are just the first thing that came to mind for each severity; you may want to come up with better ones.

With a $50k pool, that would support:

* 250 - lows
* 50 - mediums
* 10 - highs
* 2 - criticals

This seems pretty reasonable IMO. Should last at least 6 months.

come-maiz (approved):
Thanks @jtraglia.

@avalonche can you please add the levels to our bounty information? I like Justin's numbers as a way to get started, we can adjust them as we go. What about making a post in the forum with the bug bounty details and link to it in these repositories?

metachris (approved):

PR #394, opened 2022-11-04
add mev-boost version as response header to status API call

metachris:
## 📝 Summary

Adds a `X-MEVBoost-Version` header in the status API call response, which allows outside services to detect the mev-boost version.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/config/vars.go
+++ b/config/vars.go
@@ -8,6 +8,9 @@ import (
 var (
 	// Version is the version of the software, set at build time
 	Version = "v1.3.3-dev"
+
+	// ForkVersion is the latest supported fork version at build time
+	ForkVersion = "bellatrix"
 )
 
 // Other settings
--- a/server/service.go
+++ b/server/service.go
@@ -229,6 +229,7 @@ func (m *BoostService) handleRoot(w http.ResponseWriter, req *http.Request) {
 // handleStatus sends calls to the status endpoint of every relay.
 // It returns OK if at least one returned OK, and returns error otherwise.
 func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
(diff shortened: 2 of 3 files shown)


jtraglia:
Just curious, which outside services would need this? Was this a requested feature?

metachris:
CL clients might need this in the future to determine the capabilities of mev-boost (i.e. if the mev-boost version supports getPayloadV2 with Capella payload fields, etc)

jtraglia:
Oh I understand now. That makes sense. But I think I prefer the method of "try and fail" like communications with the EL client. I don't want CL clients to associate a feature with a specific version of third-party software. I see there being push back on this from client teams.

jtraglia (approved):

metachris:
> "try and fail"

yeah i second this. but it may be tricky in case of getPayload, as the CL would need to know if getPayloadV2 is available, but for that they would already send the payload at which point no fallback to local block would be reasonable 🤔 

maybe this header won't be used, but i think it can't hurt and may become useful.

jtraglia:
Ah I gotcha. Yeah that is a predicament. How about something a little more generic?

```go
w.Header().Set("SupportedMilestone", config.Milestone)
```

With values like:

* "bellatrix"
* "capella"
* "eip4844"
* ...

That would indicate support up to that milestone.

PS: There may be a better name like "fork", "version", etc.

metachris:
I like that! `ForkVersion` is used in other places, maybe `X-ForkVersion` would be a good header?

jtraglia:
Sounds good to me!

jtraglia (approved):
Thanks. If we come up with a better solution to this problem, I may recommend removing this. But for now, this works.

PR #396, opened 2022-11-04
Skip blockhash validation if version is not bellatrix

avalonche:
## 📝 Summary

 This change ignores the block hash validation for future versions of `SignedBlindedBeaconBlock` (after bellatrix)

## ⛱ Motivation and Context

For the upcoming Capella and EIP-4844 upgrade this would break block hash validation as new fields will be introduced that is not included in the block hash calculation.

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -22,10 +22,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	bellatrix = "bellatrix"
-)
-
 var (
 	errNoRelays                  = errors.New("no relays")
 	errInvalidSlot               = errors.New("invalid slot")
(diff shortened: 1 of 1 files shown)


server/service.go
```diff
@@ -543,14 +547,19 @@ func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request
...
 			// Ensure the response blockhash matches the response block
-			calculatedBlockHash, err := types.CalculateHash(responsePayload.Data)
-			if err != nil {
-				log.WithError(err).Error("could not calculate block hash")
-			} else if responsePayload.Data.BlockHash != calculatedBlockHash {
-				log.WithFields(logrus.Fields{
-					"calculatedBlockHash": calculatedBlockHash.String(),
-					"responseBlockHash":   responsePayload.Data.BlockHash.String(),
-				}).Error("responseBlockHash does not equal hash calculated from response block")
+			if responsePayload.Version == bellatrix {
```
metachris, line 550: maybe instead of hardcoding `bellatrix` use `config.ForkVersion`?

PR #398, opened 2022-11-06
Prellocate map

estensen:
## 📝 Summary

Preallocate map when we know the final size beforehand

## ⛱ Motivation and Context

Improve performance, but probably only when using many relays. 

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -314,8 +314,8 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


metachris:
Thanks for the contribution. 

There are two big problems:
1. allocating a fixed size and still using append is not working
2. the final size is not known at this time (entries to the map are only added when there's a valid bid)

Closing the PR since preallocating the map size won't work because of (2)

PR #409, opened 2022-11-15
Add support for relay weights

Agusx1211:
## 📝 Summary

Adds support for weighting relays. This affects how mev-boost determines what header to follow.

## ⛱ Motivation and Context

It allows users to evaluate multiple relays but only accept some headers when the risk/benefit outweighs the risks. For example an user could try to avoid relay X due to arbitrary reasons (censorship, trust, etc), but may want to still accept headers from it when it pays 10 times more.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -76,7 +76,7 @@ var log = logrus.NewEntry(logrus.New())
 // Main starts the mev-boost cli
 func Main() {
 	// process repeatable flags
-	flag.Var(&relays, "relay", "a single relay, can be specified multiple times")
+	flag.Var(&relays, "relay", "a single relay (optional weight can be defined using weight#relayURL), can be specified multiple times")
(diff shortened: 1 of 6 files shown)


Agusx1211:
I recognize `WEIGHT#IP@PORT` is not the cleanest schema, if that's a problem I can create a custom `wrelay` flag (or similar)

metachris:
I think this would be better started with a presentation and discussion of the idea in a Github issue or forum thread. This would allow to get a better understanding of the context and motivation as well as exploration of the mechanics.

Agusx1211:
Sure!

I just wanted the feature for myself and if I had to build it I may as well share it too. But I understand if this doesn't fit the overall vision or roadmap.

I will create a Github issue and try to explain my rationale for the idea.

Issue #404, opened 2022-11-09
Proposal: replace types from go-boost-utils with types from Attestant


metachris:
Attestant / @mcdee has been working on up-to-date types with a lot of tests: https://github.com/attestantio/go-eth2-client/tree/master/spec

I'm proposing to switch to the attestandtio types and deprecate the go-boost-utils types.

This would be one less thing to worry about, and one less repository to maintain.

ralexstokes:
this works for me, the key things to consider are:

1. ssz library used
2. maintainability, reliability of implementation

and it looks like both sources use the same ssz library so (1) is no change and I have no concerns personally about (2)

prestonvanloon:
We mentioned this in the Prysm release today. More context can be found here: https://github.com/prysmaticlabs/prysm/issues/11717

metachris:
work is underway, closing this for now.

PR #411, opened 2022-11-17
Replace go-boost-utils types with Attestant types

avalonche:
## 📝 Summary

Pr for issue #404, replacing with types from attestant to reduce maintenance for `go-boost-utils`

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cmd/test-cli/main.go
+++ b/cmd/test-cli/main.go
@@ -128,7 +128,7 @@ func doGetPayload(v validatorPrivateData, boostEndpoint string, beaconNode Beaco
 		log.WithError(err).Fatal("could not get payload")
 	}
 
-	if respPayload.Bellatrix == nil {
+	if respPayload.IsEmpty() {
 		log.Fatal("Did not receive correct payload")
 		return
 	}


cmd/test-cli/main.go
```diff
@@ -114,19 +119,19 @@ func doGetPayload(v validatorPrivateData, boostEndpoint string, beaconNode Beaco
...
+		Signature: phase0.BLSSignature(signature),
 	}
-	var respPayload boostTypes.GetPayloadResponse
+	var respPayload api.VersionedExecutionPayload
 	if _, err := server.SendHTTPRequest(context.TODO(), *http.DefaultClient, http.MethodPost, boostEndpoint+"/eth/v1/builder/blinded_blocks", "test-cli", payload, &respPayload); err != nil {
 		log.WithError(err).Fatal("could not get payload")
 	}
 
-	if respPayload.Data == nil {
+	if respPayload.Bellatrix == nil {
```
0xpanoramix, line 131: My suggestion does the same thing, but uses the method from the `go-builder-client` package.

```suggestion
	if respPayload.IsEmpty() {
```

cmd/test-cli/main.go
```diff
@@ -114,19 +119,19 @@ func doGetPayload(v validatorPrivateData, boostEndpoint string, beaconNode Beaco
...
 		log.Fatal("Did not receive correct payload")
 	}
```
0xpanoramix, line 133: Not related to the changes in your PR and not a major security issue, but if the `respPayload.Bellatrix` is nil there will be a segfault at line 134.
The following suggestion fixes the issue.

```suggestion
		log.Fatal("Did not receive correct payload")
		return
	}
```

server/service.go
```diff
@@ -345,15 +351,15 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 			}
 
 			blockHash := responsePayload.Data.Message.Header.BlockHash.String()
-			valueEth := weiBigIntToEthBigFloat(responsePayload.Data.Message.Value.BigInt())
+			valueEth := weiBigIntToEthBigFloat(responsePayload.Data.Message.Value.ToBig())
```
0xpanoramix, line 354: If for any reason the response contains a nil `Value` field, this would segfault.
This is not the case right now as `U256Str` defaults to a 32-bytes array filled with zeros.

```go

func pocGetHeaderNilMessageValueGoBoostUtils() {
	responsePayload := new(types.GetHeaderResponse)
	responsePayload.Data = &types.SignedBuilderBid{
		Message: &types.BuilderBid{},
	}

	fmt.Println(responsePayload.Data.Message.Value)
}

func pocGetHeaderNilMessageValueAttestantTypes() {
	responsePayload := new(spec.VersionedSignedBuilderBid)
	responsePayload.Data = &bellatrix.SignedBuilderBid{
		Message: &bellatrix.BuilderBid{},
	}

	fmt.Println(responsePayload.Data.Message.Value)
}

func main() {
	// Prints the following:
	// [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
	// <nil>

	pocGetHeaderNilMessageValueGoBoostUtils()
	pocGetHeaderNilMessageValueAttestantTypes()
}
```

What I would suggest (I can create one here because the corresponding lines were not edited) is to add a `.Value != nil` in the if statement at line 349.

server/utils.go
```diff
@@ -142,3 +148,49 @@ func weiBigIntToEthBigFloat(wei *big.Int) (ethValue *big.Float) {
...
+	}
+	return phase0.Hash32(header.Hash()), nil
+}
+
+func executionPayloadToBlockHeader(payload *bellatrix.ExecutionPayload) (*types.Header, error) {
+	transactionData := make([]*types.Transaction, len(payload.Transactions))
+	for i, encTx := range payload.Transactions {
+		var tx types.Transaction
+		err := tx.UnmarshalBinary(encTx)
+		if err != nil {
```
0xpanoramix, line 165: Just a little styling comment.

```suggestion

		if err := tx.UnmarshalBinary(encTx); err != nil {
```

server/utils.go
```diff
@@ -142,3 +148,49 @@ func weiBigIntToEthBigFloat(wei *big.Int) (ethValue *big.Float) {
...
+			return nil, errInvalidTransaction
+		}
+		transactionData[i] = &tx
+	}
+
+	// base fee per gas is stored little-endian but we need it
+	// big-endian for big.Int.
+	var baseFeePerGasBytes [32]byte
+	for i := 0; i < 32; i++ {
+		baseFeePerGasBytes[i] = payload.BaseFeePerGas[32-1-i]
```
0xpanoramix, line 175:
```suggestion
		baseFeePerGasBytes[i] = payload.BaseFeePerGas[31-i]
```

0xpanoramix (changes requested):
Nice improvements !
I added a few comments about potential missing checks and styling :)

server/utils.go
0xpanoramix: I see, lets keep this way then.

server/service.go
0xpanoramix: Thanks for the link, I did not realized it was already verified 👍

metachris:
Really nice work! Keeping this open for a bit longer for more review, but great start on moving towards Capella&EIP4844 compatibility.

One thing to notice, if this is merged and proposers would upgrade mev-boost but not Prysm, then it might lead to those users missing their slot. See also:

* https://github.com/prysmaticlabs/prysm/issues/11717

metachris:
The next step here would be to 
1. decode into Capella(Attestant) types 
2. fall-back to Bellatrix(go-boost-utils) types

This will make these changes fully backwards-compatible, without causing any possible issues with Prysm users.

Seyvick (approved):

Seyvick:
Aprobado

Issue #412, opened 2022-11-17
Should "latest" Docker image be the portable image (it is since last release)?


metachris:
Currently the Docker CI setup is unclear, and the portable build-and-push step is overwriting the `latest` image with the portable image.

Do we want the latest image to be the portable image?

* If yes, then we should disable the `latest` tag for the non-portable image build step
* If no, then we should disable the `latest` tag for the portable image build step

metachris:
closing, latest is the portable build, starting at mev-boost v1.4.0

PR #413, opened 2022-11-17
fix CI latest Docker image tagging

metachris:
## 📝 Summary

Fix CI overwriting the `latest` Docker image with both portable and non-portable images (as seen here in the latest release CI action: https://github.com/flashbots/mev-boost/actions/runs/3428613593/jobs/5713128560)

Currently, the `latest` image on Docker hub is the portable image.

This PR updates the Docker release CI step to do that explicitly.

Also removes pushing images for `:1` and `:1.4` when releasing `v1.4.0` (these image tags seem unnecessary and I'd opt for removing them from Docker hub, otherwise we should also add them for portable).

See also #412 

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,9 +33,7 @@ jobs:
           tags: |
             type=sha
             type=semver,pattern={{version}}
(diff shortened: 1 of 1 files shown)


jtraglia (approved):
LGTM 👍

PR #423, opened 2022-11-28
Change vars to const

estensen:
## 📝 Summary

Extract some vars to const

## ⛱ Motivation and Context

const vals that shouldn't change

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/mock_relay.go
+++ b/server/mock_relay.go
@@ -19,7 +19,6 @@ import (
 
 const (
 	mockRelaySecretKeyHex = "0x4e343a647c5a5c44d76c2c58b63f02cdf3a9a0ec40f102ebc26363b4b1b95033"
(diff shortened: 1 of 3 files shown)


server/mock_relay.go
```diff
@@ -17,9 +17,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var (
+const (
 	mockRelaySecretKeyHex = "0x4e343a647c5a5c44d76c2c58b63f02cdf3a9a0ec40f102ebc26363b4b1b95033"
 	// mockRelayPublicKeyHex = "0x8a1d7b8dd64e0aafe7ea7b6c95065c9364cf99d38470c12ee807d55f7de1529ad29ce2c422e0b65e3d5a05c02caca249"
```
ralexstokes, line 22: a bit off-topic but if we don't need this comment, can we just remove it?

estensen: Removed it

metachris (approved):

metachris:
thanks 🙏

Seyvick (commented):
.

PR #424, opened 2022-11-29
Add -min-bid 0.05 to strengthen censorship resilience

sandakersmann:
Add -min-bid 0.05 to strengthen censorship resilience i.a.w: https://writings.flashbots.net/the-cost-of-resilience/

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ Type=simple
 Restart=always
 RestartSec=5
 ExecStart=/home/mev-boost/bin/mev-boost \
+        -min-bid 0.05 \
         -relay-check \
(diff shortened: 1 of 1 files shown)


ralexstokes:
in the spirit of `mev-boost` itself being neutral software, I don't think we want to set any prescriptive norms around how the software should be used

others in the community should feel free to make recommendations/suggestions around using the software but that's out of scope for this repo. i'll close this PR with this reasoning

sandakersmann:
People usually run software with the defaults. I don't see how suggesting defaults that aligns with the values of Ethereum is not a good thing. This should be merged.

PR #440, opened 2023-01-25
Drain body completely to reuse connection

estensen:
## 📝 Summary

Drain body completely to reuse connection

## ⛱ Motivation and Context

Reuse connections when possible

	// non-nil, even on responses without a body or responses with
	// a zero-length body. **It is the caller's responsibility to
	// close Body. The default HTTP client's Transport may not
	// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
	// not read to completion and closed.**
https://pkg.go.dev/net/http#Response

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -63,7 +63,11 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
 	if err != nil {
 		return 0, err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		// drain body completely to reuse connection
+		_, _ = io.Copy(io.Discard, resp.Body)
+		resp.Body.Close()
+	}()
 
 	if resp.StatusCode == http.StatusNoContent {
 		return resp.StatusCode, nil


server/utils.go
```diff
@@ -63,7 +63,11 @@ func SendHTTPRequest(ctx context.Context, client http.Client, method, url string
 	if err != nil {
 		return 0, err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		// drain body completely to reuse connection
+		_, _ = io.Copy(io.Discard, resp.Body)
```
ralexstokes, line 68: do you have any data on a lack of reuse of connections?

if you look below the `io.ReadAll` utility is used to consume the response body so according to the docs, the `Close` (as currently exists) is sufficient to trigger the reuse mechanism

metachris: Thanks Alex, would be curious too.

metachris:
Thanks for the contribution! Do you have a reference for this being a best practice? I'm not sure there is a problem in the current impl 🤔

metachris:
Closing this for now. Please comment with references/examples and we can reopen! 🙏

PR #454, opened 2023-02-21
use Go v1.20

metachris:
## 📝 Summary

Go v1.20.1 is the latest. Default to 1.20 for building and testing.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,10 +16,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Set up Go 1.19
+      - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19
+          go-version: ^1.20
           id: go
 
       - name: Ensure go mod tidy runs without changes
@@ -29,13 +29,13 @@ jobs:
           git diff-index --quiet HEAD
 
       - name: Install gofumpt
-        run: go install mvdan.cc/gofumpt@v0.3.1
+        run: go install mvdan.cc/gofumpt@v0.4.0
 
       - name: Install staticcheck
-        run: go install honnef.co/go/tools/cmd/staticcheck@v0.3.3
+        run: go install honnef.co/go/tools/cmd/staticcheck@v0.4.2
 
       - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.49.0
+        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2
 
       - name: Lint
         run: make lint
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-FROM golang:1.19 as builder
+FROM golang:1.20 as builder
 ARG VERSION
 ARG CGO_CFLAGS
 WORKDIR /build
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY go.sum ./
 RUN go mod download
 
 ADD . .
-RUN --mount=type=cache,target=/root/.cache/go-build CGO_CFLAGS="$CGO_CFLAGS" GOOS=linux go build -trimpath -ldflags "-s -X 'github.com/flashbots/mev-boost/config.Version=$VERSION'" -v -o mev-boost .
+RUN --mount=type=cache,target=/root/.cache/go-build CGO_CFLAGS="$CGO_CFLAGS" GOOS=linux go build -trimpath -ldflags "-s -X 'github.com/flashbots/mev-boost/config.Version=$VERSION' -linkmode external -extldflags '-static'" -v -o mev-boost .
```
metachris, line 13: Needed to compile fully static binary. Otherwise getting the following error when running the Docker image:

```
Error loading shared library libresolv.so.2: No such file or directory (needed by /app/mev-boost)
```

jtraglia: This seems fine to me.

.github/workflows/lint.yml
```diff
@@ -16,10 +16,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Set up Go 1.19
+      - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19
+          go-version: ^1.20
```
jtraglia, line 22: Just to be clear, this is the same as `1.20.x` (any patch version) right?

Had to check the documentation:
* https://github.com/npm/node-semver#caret-ranges-123-025-004

server/service_test.go
```diff
@@ -182,18 +182,18 @@ func TestWebserverMaxHeaderSize(t *testing.T) {
...
+// 		FeeRecipient: _HexToAddress("0xdb65fEd33dc262Fe09D9a2Ba8F80b329BA25f941"),
+// 		Timestamp:    1234356,
+// 		GasLimit:     278234191203,
+// 		Pubkey: _HexToPubkey(
+// 			"0x8a1d7b8dd64e0aafe7ea7b6c95065c9364cf99d38470c12ee807d55f7de1529ad29ce2c422e0b65e3d5a05c02caca249"),
+// 	},
+// 	// Signed by 0x4e343a647c5a5c44d76c2c58b63f02cdf3a9a0ec40f102ebc26363b4b1b95033
+// 	Signature: _HexToSignature(
+// 		"0x8209b5391cd69f392b1f02dbc03bab61f574bb6bb54bf87b59e2a85bdc0756f7db6a71ce1b41b727a1f46ccc77b213bf0df1426177b5b29926b39956114421eaa36ec4602969f6f6370a44de44a6bce6dae2136e5fb594cce2a476354264d1ea"),
+// }
```
jtraglia, line 196: Why did this get commented out?

.github/workflows/lint.yml
metachris: yes! same as in go.mod and the Dockerfile 👍

server/service_test.go
jtraglia: Yeah if it's not being used we should remove it.

metachris: done

jtraglia (approved):
LGTM 👍

PR #465, opened 2023-03-09
getPayload retry with 100ms timeout

metachris:
## 📝 Summary

Wait 100ms until getPayload retry for higher success rate

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


Ruteri (approved):
I think a sleep like this warrants a comment so that we won't get a PR saying "removed the 100ms sleep, performance is 10% better now"

metachris:
added the comment!

jtraglia (approved):
LGTM 👍

Issue #455, opened 2023-02-21
Dynamic Relay Configuration Proposal


screwyprof:
# Dynamic Relay Configuration Proposal

This proposal summarises our take on[ Support Validator Account Level Relay Choice](https://github.com/flashbots/mev-boost/issues/154) and [Validator Level Relay Configuration](https://github.com/flashbots/mev-boost/issues/416) as part of ConsenSys Codefi team responding to the features our customers requested.

## Requirements

Many stakers run multiple keys on the same validator/consensus layer client. **MEV-Boost**, as currently implemented, requires all keys on the same node to point at the same set of relays.

<img alt="meg-boost before" src="https://user-images.githubusercontent.com/1536274/220412386-d13661e3-81b4-4e3d-ae0a-99a505d045ad.png" width="400" />

The objective of this proposal is to allow stakers to specify different relays for different validators without having to increase their consensus layer client population: 

<img alt="mev-boost after" src="https://user-images.githubusercontent.com/1536274/220412381-2e71c192-ef20-4c04-8beb-4a3f8e1fa7d9.png" width="400" />

## Acceptance Criteria

- Different validator keys on the same validator client can point to different sets of relays
- There is no compromise on security or introduction of a new attack surface
- Minimal impact on the current latency sensitive operations of **MEV-Boost**
- The implementation is optional - users can continue to use **MEV-Boost** as they do today
- Changes to the relay set of a particular validator are adopted quickly by the infrastructure with no disruption (e.g.: restarting **MEV-Boost**)

## Definitions

**Relay Configuration Provide**r (**RCP**) holds information about validators and their corresponding relays. For example it may be as simple as a file or a remote API endpoint which returns a [Proposer Config](https://docs.teku.consensys.net/HowTo/Configure/Proposer-Configuration) - a JSON discussed in [Support validator account level relay choice](https://github.com/flashbots/mev-boost/issues/154) issue.

**Relay Configuration Manager (RCM)** manages relays and their configurations. It fetches the configuration from an **RCP**, validates it and then populates an in memory **Relay Registry** which is basically a map of all the registered validators and their corresponding relay configs. The **Relay Registry** is then used to lookup relays for a given validator when requested. 

## Suggested Changes

We change **MEV-Boost** to:

1. Support a command line/env config flag to use an **RCM**.
2. Call the **RCM** to fetch all the relays for all the managed validators at startup.
3. Query the **RCM** to get relays for a given validator on every **MEV-Boost** API call.

[...]

PR #470, opened 2023-03-14
Dynamic relay configuration (fixed)

screwyprof:
## 📝 Summary

This PR implements [Dynamic Relay Configuration Proposal](https://github.com/flashbots/mev-boost/issues/455).

At the moment 2 RCPs are implemented: 
- File
- JSON API

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


cli/main.go
```diff
@@ -225,6 +250,118 @@ func Main() {
...
+	return nil
+}
+
+func mapKeysToString(m map[string]string) string {
+	res := make([]string, 0, len(m))
+	for opt := range m {
+		res = append(res, opt)
+	}
+
+	return "-" + strings.Join(res, " or -")
```
jtraglia, line 299: For a consistent error message:
```suggestion
	sort.Strings(res)
	return "-" + strings.Join(res, " or -")
```

Without this they will show up in a random order. Also, I would prefer they were in this order:

1. `-relay`
2. `-relays`
3. `-proposer-config-file`
4. `-proposer-config-url`

But that would be more complicated & I don't believe it's worth it.

config/rcm/configurator.go
```diff
@@ -0,0 +1,76 @@
...
+
+// AllRelays retrieves a list of all unique relays from the Relay Registry.
+func (m *Configurator) AllRelays() relay.List {
+	return m.loadRegistry().AllRelays()
+}
+
+func (m *Configurator) loadRegistry() RelayRegistry {
+	r, ok := m.relayRegistry.Load().(RelayRegistry)
+	if !ok {
+		panic("unexpected relay registry type") // this must never happen
```
jtraglia, line 72: You can remove the "this must never happen" comment. Doesn't seem that useful.

config/rcm/sync_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/flashbots/mev-boost/config/rcp/rcptest"
+	"github.com/flashbots/mev-boost/config/relay"
+	"github.com/flashbots/mev-boost/config/relay/reltest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
```
jtraglia, line 20: That's a neat package.

server/service.go
```diff
@@ -610,9 +665,9 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
 	requestCtx, requestCtxCancel := context.WithCancel(context.Background())
 	defer requestCtxCancel()
 
-	for _, relay := range relays {
+	for _, r := range relays {
```
jtraglia, line 668: Guessing this change was made because the inner function used the same name? Good change.

server/service.go
```diff
@@ -484,7 +537,8 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
 	relays := originalBid.relays
 	if len(relays) == 0 {
 		log.Warn("originating relay not found, sending getPayload request to all relays")
-		relays = m.relays
+
+		relays = m.relayConfigurator.AllRelays()
```
jtraglia, line 541: Can we remove some of the newly-added blank lines in this file? To me it looks wrong.

jtraglia (commented):
Here's my initial review. For the most part, this PR looks really good. I recognize the problem and this seems like a good solution. I like the coding style, appreciate all of the tests/comments, and think it's organized well. I didn't notice any problems, but I'm concerned with just how much has changed. There's a _lot_ of new code and modifications. I'm not sure that I would ever feel 100% confident with this PR, but that's okay. I'd say, as long as mev-boost remains backwards-compatible for existing users and it doesn't introduce any new vulnerabilities, this PR will eventually have my blessing.

config/rcm/configurator.go
screwyprof: I wish Golang had something like `unreachable!` in rust. This comment justifies using a panic here.

server/service.go
screwyprof: `relay` collides with the package name, hence the change.

server/service.go
screwyprof: Not really, otherwise the linters will complain.  Statements and assignments must not be coupled.

server/service.go
jtraglia: Ah gotcha. Makes sense.

server/service.go
jtraglia: Which linter complains about that?

I don't see any complains about that in `make lint` with these changes:

<details>
<summary>Diff</summary>

```diff
diff --git a/server/service.go b/server/service.go
index 1ffbfb0..0841798 100644
--- a/server/service.go
+++ b/server/service.go
@@ -295,7 +295,6 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 		if len(relays) < 1 {
 			log.Warnf("there are no relays specified for %s", pubKey)
 		}
-
 		relayPayloads.add(relays, p)
 	}
 
@@ -307,15 +306,11 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 
 		go func(r relay.Entry, payloads []types.SignedValidatorRegistration) {
 			defer wg.Done()
-
 			relayURL := r.GetURI(pathRegisterValidator)
-
 			_, err := SendHTTPRequest(context.Background(), m.httpClientRegVal, http.MethodPost, relayURL, ua, payloads, nil)
 			relayErrCh <- err
-
 			if err != nil {
 				log.WithField("url", relayURL).WithError(err).Warn("error calling registerValidator on relay")
-
 				return
 			}
 		}(r, payloads)
@@ -331,7 +326,6 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 	for err := range relayErrCh {
 		if err == nil {
 			m.respondOK(w, nilResponse)
-
 			return
 		}
 	}
@@ -543,7 +537,6 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
 	relays := originalBid.relays
 	if len(relays) == 0 {
 		log.Warn("originating relay not found, sending getPayload request to all relays")
-
 		relays = m.relayConfigurator.AllRelays()
 	}
 
@@ -558,7 +551,6 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
 
 	for _, r := range relays {
 		wg.Add(1)
-
 		go func(relay relay.Entry) {
 			defer wg.Done()
 			url := relay.GetURI(pathGetPayload)
@@ -778,7 +770,6 @@ func (m *BoostService) CheckRelays() int {
 
 	for _, r := range m.relayConfigurator.AllRelays() {
 		wg.Add(1)
-
 		go func(relay relay.Entry) {
 			defer wg.Done()
 			url := relay.GetURI(pathStatus)
```
</details>

config/rcm/configurator.go
jtraglia: I think `panic` implies that it should never happen.

server/service.go
screwyprof: Sorry for confusion. I was running a much stricter set of linters locally on the newly introduced changes via pre-commit hook. `wsl` was included.

screwyprof: Also could me tell me if the project has a style guide where it's recommended to no use blank lines?

config/rcm/configurator.go
screwyprof: I'd argue panic just implies "panic" unless you have a context or an agreement to say otherwise. That's why I added the comment. Happy to fix if you really think we should do it.

server/service.go
jtraglia: There is no well-defined style guide, it just needs to pass the formatting checks:

https://github.com/flashbots/mev-boost/blob/3b7f08b6e2c38730b7fc37edc73b147a1d5df0c8/Makefile#L56-L58

Generally, this project leans towards denser code with less whitespace.

jtraglia: I'm not necessarily against that style, it's just a little inconsistent with the other functions in that file.

screwyprof: The code passed the checks, didn't it? I'll remove the line to save us some time for some more important things than a holy war on blank lines :)

config/rcm/configurator.go
metachris: yes agree, panic implies it should never happen, comment seems not needed

server/service.go
metachris: what's the case for adding a bunch of new newlines? that's generally not considered best practice, i'd say :)

screwyprof: That’s a strong assertion, as some people and linters will disagree. However if you want me to remove the blank lines, I will do that. As I said, with no style guide in place this is just a matter of taste. Please let me know if there are any other points that should be addressed as I’d rather fix them all at once.

cli/main.go
screwyprof: Fixed.

config/rcm/configurator.go
screwyprof: Fixed.

server/service.go
screwyprof: Fixed, empty lines removed.

metachris:
Just a quick heads-up: this PR is still on the radar, and as soon as things around Capella settle down (mainnet upgrade on April 12), we'll have enough resources to focus on this properly.

screwyprof:
Hey guys, It's been a while since Capella/Shanghai were merged, any news or ETA?

metachris:
Hey! It's just been a few days -- Shapella just really wrapped up Thursday night. Will get to it soon.

server/service.go
```diff
@@ -243,44 +250,75 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+type (
```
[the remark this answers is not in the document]
metachris: or at least have a comment that these are registerValidator specific types

server/service.go
```diff
@@ -243,44 +250,75 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
 
-	for _, relay := range m.relays {
-		go func(relay RelayEntry) {
-			url := relay.GetURI(pathRegisterValidator)
-			log := log.WithField("url", url)
+	relayPayloads := make(payloadsByRelay, len(m.relayConfigurator.AllRelays()))
+	for pubKey, p := range validatorPayloads {
+		relays := m.relayConfigurator.RelaysForProposer(pubKey)
+		if len(relays) < 1 {
+			log.Warnf("there are no relays specified for %s", pubKey)
```
metachris, line 290: could add a `continue` here after the log

server/service.go
```diff
@@ -243,44 +250,75 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
+		}
+		relayPayloads.add(relays, p)
+	}
 
-			_, err := SendHTTPRequest(context.Background(), m.httpClientRegVal, http.MethodPost, url, ua, payload, nil)
-			relayRespCh <- err
+	wg := new(sync.WaitGroup)
```
metachris, line 295: i prefer the scatter-gather pattern that we used to have, over this combination of waitgroup and errChan

server/service.go
```diff
@@ -243,44 +250,75 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
-		respErr := <-relayRespCh
-		if respErr == nil {
+	go func() {
+		wg.Wait()
+		close(relayErrCh)
+	}()
+
+	for err := range relayErrCh {
+		if err == nil {
 			m.respondOK(w, nilResponse)
```
metachris, line 322: so this returns OK on the first successful response from any relay.

does this affect the other requests? should it wait for all to complete?

server/service.go
```diff
@@ -243,44 +250,75 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
+	for _, p := range payload {
+		pubKey := p.Message.Pubkey.String()
+		validatorPayloads[pubKey] = p
+	}
 
-	for _, relay := range m.relays {
-		go func(relay RelayEntry) {
-			url := relay.GetURI(pathRegisterValidator)
-			log := log.WithField("url", url)
+	relayPayloads := make(payloadsByRelay, len(m.relayConfigurator.AllRelays()))
```
metachris, line 286: I think this should not preallocate the map size, but only add when they're actually used.

that way things can be a bit cleaner, and use the scatter-gather concurrency pattern as in the original codebase

server/service.go
```diff
@@ -243,44 +250,75 @@ func (m *BoostService) handleStatus(w http.ResponseWriter, req *http.Request) {
...
 	ua := UserAgent(req.Header.Get("User-Agent"))
-	log = log.WithFields(logrus.Fields{
+	log.WithFields(logrus.Fields{
 		"numRegistrations": len(payload),
 		"ua":               ua,
-	})
+	}).Info("registering validators")
 
-	relayRespCh := make(chan error, len(m.relays))
+	validatorPayloads := make(payloadByValidator, len(payload))
```
metachris, line 280: no need to preallocate map size. could lead to weird results if for some reason a CL client includes multiple registrations for a single validator

server/service.go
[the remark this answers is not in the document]
screwyprof: Fixed.

server/service.go
screwyprof: Fixed

server/service.go
screwyprof: The behaviour is exactly the same as it was before the changes.

server/service.go
screwyprof: Fixed.

server/service.go
screwyprof: Fixed

metachris:
Hey Max, time for an update!

I had this PR on the back of my mind for a while, and did a few partial reviews already.

Then there was a lot of work due to the unbundling attack, and the Shapella upgrade in the week after.

Generally the PR looks good, but I am having a hard time with because of it’s size, and the depth of the changes. That makes it very hard to reason about and ensuring it’s correctness, and it's very important to be careful with merging these things.

For adding such an amount of changes, it would be good to get review and sign-off from multiple core contributors, who I believe are running into the same problem when reviewing this PR - that it’s hard to verify because of it’s size.

My suggestion would be to break up this PR into multiple smaller ones. For instance, start with only the container, or whatever you think would be a good first step. Do you have an idea how it could be broken up effectively? That way it will be easier to review each of the PRs thoroughly and can move this forward.

ralexstokes:
I think we all agree more flexible relay configuration is a very nice enhancement to this software; that being said, I need to echo metachris that this software has become quite important to the Ethereum ecosystem and so changes need to be merged carefully.

I hear you on the back and forth and understand how frustrating that can be. But I do find it hard to review something this large and so some suggestions:

* I see some whitespace edits, can we pull these out into a separate PR?
* Can we start with just the core changes in a single PR? Which seem to be the `Configurator` and which every implementation is the simplest (I think just the config file provider, right?)

Once we merge in the skeleton, we can add fancier features like additional configuration providers that pull in more concepts like an API provider

beetrootkid:
Hi @metachris and @ralexstokes . Thanks a lot for your additional feedback and pointers. We've had a chance to discuss this now internally and wanted to follow up on a couple of things:

1. Can you and the team confirm that you actually want this feature in mev-boost? Regardless of how the PR is sliced and diced, it would be good if you could confirm that this is something you want to and will add.
2. Given that we've provided very detailed and clear documentation, is it possible for you to suggest how exactly you'd like the PR to be split? I know @ralexstokes you've made some suggestions above. Could you add a little more detail and perhaps an indication of how many PRs you'd think we should aim for?

Thanks a lot.

metachris:
@beetrootkid

Re 1 - i think the question is more whether there is broader interest in this feature. i.e. have other staking pools expressed interest? Feels like this functionality could be useful to many, but i haven't seen that expressed. I'm not used and quite positive about this functionality generally speaking, even though it introduces significant complexity.

Re 2 - there already were some suggestions how to break it up, and i think as developers you could make the most educated suggestion about this. Anything from 2 to 4 PRs sound like a reasonable number.

Alex is out this week. Pinging @jtraglia and @michaelneuder for thoughts

metachris:
this PR was closed automatically (and by accident) when removing the `main` branch, sorry about that!

metachris:
a

metachris (changes requested):
As discussed in the comments, please break this up into smaller changes 🙏

metachris:
Closing this for now since the last comments were over a month ago. Happy to reopen / continue the conversation if there's interest @beetrootkid. Thanks for the effort

PR #474, opened 2023-03-16
fix: goerli/prater genesis validator root

nick-staked:
It looks like the genesis hash was used by mistake: https://github.com/eth-clients/goerli/#meta-data-prater

## 📚 References

---

* [ ] `make lint`
* [ ] `make test-race`
* [ ] `go mod tidy`


metachris (approved):

metachris:
thanks, i think that's correct. @Ruteri wdyt?

PR #479, opened 2023-03-17
Replace blst with gnark-crypto, simplify builds

metachris:
## 📝 Summary

Replacing [blst](https://github.com/supranational/blst) with [gnark-crypto](https://github.com/ConsenSys/gnark-crypto), an audited, non-cgo BLS12-381 implementation.

Follow-up to initial work from @jtraglia at

- https://github.com/flashbots/go-boost-utils/pull/56
- https://github.com/flashbots/go-boost-utils/pull/62

No more portable/static builds needed either! Non-cgo builds are static by default.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


.goreleaser-darwin.yaml
```diff
@@ -11,8 +11,6 @@ builds:
       - -X github.com/flashbots/mev-boost/config.Version={{.Version}}
     env:
       - CGO_ENABLED=1
```
jtraglia, line 13: I think we should hardcode this to `CGO_ENABLED=0` now. This would further simplify things. With cgo disabled, go builds are static by default. We'd also be able to 

* remove `osusergo` and `netgo` build tags,
* remove `-extldflags=-static` configuration,
* remove `CC=<compiler>` environment variables,
* remove the `build-static` makefile rule,
* remove the `EXTRA_ENV` makefile variable, and
* join ` GO_BUILD_LDFLAGS` into `GO_BUILD_FLAGS`.

metachris: Great suggestion! Added a commit that does part of the cleanup. 

Would be amazing if you could combine the `gorelease-*.yml` files in another PR 🥺

jtraglia: Nice 👍 yes, I can clean up the releases in another PR.

jtraglia:
I pushed a few commits that explicitly disabled cgo in more places (like tests). It needs to be enabled with the `-race` flag though, but I think that's fine. Also, one extra change is the removal of `-v` (verbose) from the builds; I don't think it's all that useful anymore. I think we should have another review from someone else before merging this.

Makefile
```diff
@@ -82,19 +67,10 @@ run-mergemock-integration: build
...
 	docker tag mev-boost:latest ${DOCKER_REPO}:latest
 
-.PHONY: docker-image-portable
-docker-image-portable:
-	DOCKER_BUILDKIT=1 docker build --platform linux/amd64 --build-arg CGO_CFLAGS="-O -D__BLST_PORTABLE__" --build-arg VERSION=${VERSION}  . -t mev-boost
-	docker tag mev-boost:latest ${DOCKER_REPO}:${VERSION}
-
-.PHONY: docker-push-version
-docker-push-version:
-	docker push ${DOCKER_REPO}:${VERSION}
```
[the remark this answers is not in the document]
metachris: yep, better to remove the manual steps, which i haven't done in a long time, and rely on github CI for pushing docker images. if we'd keep it, it should also push the latest tag. i think better remove for now.

jtraglia:
Please wait to merge this until we fully consider the security/performance implications.

* IIRC, `kilic` has **not** been audited, but neither has `go-ethereum/crypto/bls12381`.
  * Would we still feel safer using geth's implementation, which is based off an old version of kilic? 
* Do the EL/CL clients double check critical signatures?
  * They wouldn't check relay signatures, is that critical?
* How much of a performance penalty does this cost?
  * I fully expect `blst` to be faster, but is that worth the extra complexity?

metachris:
Thanks, good considerations! 

I like the PR https://github.com/flashbots/go-boost-utils/pull/62 to replace it with https://github.com/ConsenSys/gnark-crypto -- seems like a great replacement, and should certainly be an improvement over https://github.com/supranational/blst

> They wouldn't check relay signatures, is that critical?

It's not, that was a responsibility solely on mev-boost side.

go.mod
```diff
@@ -19,6 +19,8 @@ require (
 	github.com/cockroachdb/logtags v0.0.0-20230118201751-21c54148d20b // indirect
 	github.com/cockroachdb/pebble v0.0.0-20230209160836-829675f94811 // indirect
 	github.com/cockroachdb/redact v1.1.3 // indirect
+	github.com/consensys/bavard v0.1.13 // indirect
+	github.com/consensys/gnark-crypto v0.9.1-0.20230105202408-1a7a29904a7c // indirect
```
jtraglia, line 23: Why is this using a different version of gnark-crypto than go-boost-utils? It should be `v0.7.1`.

jtraglia:
@metachris Have you double-checked that the `goreleaser` changes are good?

jtraglia:
Updating [`go-ole`](https://pkg.go.dev/github.com/go-ole/go-ole@v1.2.6) to the latest version (`v1.2.6`) files the Windows/ARM64 build issue.

The errors from previous version:

```
$ GOOS=windows GOARCH=arm64 go build
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/com.go:238:21: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/com.go:247:22: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:79:72: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:90:76: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:105:69: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:115:73: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:129:69: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:139:73: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/connect.go:173:84: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/idispatch.go:26:90: undefined: VARIANT
../../../go/pkg/mod/github.com/go-ole/go-ole@v1.2.5/idispatch.go:26:90: too many errors
```

go.mod
jtraglia: Ah, it's because Geth [uses gnark-crypto](https://github.com/ethereum/go-ethereum/blob/master/go.mod#L16) for two of their fuzzers:

* [`bn256_fuzz.go`](https://github.com/ethereum/go-ethereum/blob/e6b6a8b738069ad0579f6798ee59fde93ed13b43/tests/fuzzers/bn256/bn256_fuzz.go#L28)
* [`bls12381_fuzz.go`](https://github.com/ethereum/go-ethereum/blob/e6b6a8b738069ad0579f6798ee59fde93ed13b43/tests/fuzzers/bls12381/bls12381_fuzz.go#L29)

And because this is just for fuzzers, it seems perfectly fine to use the latest version.

I'm fairly certain that Go is smart enough to handle two version of indirect dependencies.

jtraglia: Hmm actually, I think that mev-boost is using `v0.9.1`, not `v0.7.1` that go-boost-utils requires. If true, that's pretty disappointing and feels like a Go problem. If you delete the module cache (`go clean -modcache`) and rebuild, it only downloads `v0.9.1`. And I just realized that go-boost-utils works with both versions, so the tests aren't an indicator.

jtraglia: Fixed with a `replace` in the `go.mod` file.

metachris:
very cool!

btw, added a commit to ignore `gomoddirectives`, which makes the linter work again

metachris:
created a release based on this branch and state, to test in testnets before merging:

* https://github.com/flashbots/mev-boost/releases/tag/v1.5.2-alpha2
* [Docker image](https://hub.docker.com/r/flashbots/mev-boost/tags): `flashbots/mev-boost:1.5.2-alpha2`

metachris:
Asked @parithosh to help test in some of the testnets

jtraglia:
> btw, added a commit to ignore `gomoddirectives`, which makes the linter work again

Ah thank you, I didn't notice that the linter failed CI checks. I just made a change to your fix. Instead of disabling that linter check, add `gnark-crypto` to the `replace-allow-list` so that specific one is allowed.

Also, I've checked out the release and it looks good! Running `1.5.2-alpha2` via docker with my setup now.

jtraglia:
Sorry to flip-flop again, but I think we should be using the latest version of `gnark-crypto` instead of the audited version. I think it's beneficial to get all of the minor improvements & bug fixes that the newer versions include.

Updated this PR accordingly.

metachris:
Thanks for keeping an eye on the details! <3

jtraglia (approved):
I'm a bit biased, but LGTM 👍 I'm happy with this PR!

PR #493, opened 2023-04-18
Do not allow relays to have point-at-infinity pubkey

jtraglia:
## 📝 Summary

* Return an error if a relay is configured with the point-at-infinity (all-zero) pubkey.
* Delete blank lines in test cases (just a nit picky thing I want to fix).
* Rename a test case for consistency:
  * From: "Invalid relay public key"
  * To: "Relay URL with invalid public key"

## ⛱ Motivation and Context

@asanso pointed out that if a relay were configured with the point-at-infinity public key, it could essentially bypass signature verification checks when receiving bids. In practice, this isn't a problem. A user would need to configure mev-boost with this and there's already a trusted relationship with relays (_i.e.,_ we trust them to only send valid bids/payloads). But I do think it's a good situation to avoid and that's why I'm making this PR.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/relay_entry.go
+++ b/server/relay_entry.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"bytes"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


server/relay_entry.go
```diff
@@ -41,8 +42,19 @@ func NewRelayEntry(relayURL string) (entry RelayEntry, err error) {
...
 
+	// Convert the username string to a public key.
 	err = entry.PublicKey.UnmarshalText([]byte(entry.URL.User.Username()))
-	return entry, err
+	if err != nil {
+		return entry, err
+	}
+
+	// Check if the public key is the point-at-infinity.
+	pointAtInfinity := [48]byte{}
```
ralexstokes, line 52: worth having this as a constant to avoid allocating each call?

ralexstokes (approved):
generally looks good, I like reducing attack surface here, even though I agree this is a bit minor

server/relay_entry.go
jtraglia: Sure, I'm good with that. Fixed.

metachris (approved):

metachris:
thanks!

PR #494, opened 2023-04-19
only run github ci checks once

metachris:
## 📝 Summary

Before they were always run twice (once for push and once for PR). This PR makes the checks run only once.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,8 +2,9 @@ name: Tests
 
 on:
   push:
+    branches:
+      - main
   pull_request:
-    branches: [main]
 
 jobs:
   test:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,8 +2,9 @@ name: Linting
 
 on:
   push:
+    branches:
+      - main
   pull_request:
-    branches: [main]
 
 jobs:
   lint:


.github/workflows/tests.yml
```diff
@@ -2,8 +2,9 @@ name: Tests
 
 on:
   push:
+    branches:
+      - main
   pull_request:
```
jtraglia, line 7: Nit: we can remove this section now, right? (for both of the files)
```suggestion
```

jtraglia (approved):
LGTM 👍 just one nit.

jtraglia:
Though, are you sure this is the desired behavior? Without this, I don't believe it will run CI checks for PRs. And I think it's pretty useful to see the CI checks for in PRs.

ralexstokes:
yeah if anything I'd rather enforce the checks at the PR, then not bother with running them "twice" on the main branch

that being said I am also curious why we need to change from the status quo in the first place?

.github/workflows/tests.yml
metachris: we cannot remove this section, because then the checks wouldn't run on PRs!

metachris:
We need to change from status quo because in the current configuration, checks are only run for PRs to the main branch, but they should be run on any PR (i.e. you open a PR to another PR/another branch).

The new configuration is correct -- run checks on every PR, and on any push to main branch too

jtraglia:
Okay, sounds good to me 👍

PR #496, opened 2023-04-21
Replace go-boost-utils types with attestant types

jtraglia:
## 📝 Summary

Replace go-boost-utils types with [`attestantio/go-builder-client`](https://github.com/attestantio/go-builder-client) and [`attestantio/go-eth2-client`](https://github.com/attestantio/go-eth2-client) types.

* Replace types with Attestant types.
* Uncomment out a few test cases.
* Add some missing field to the `kiln-signed-blinded-beacon-block` which were missing.
  * The go-boost-utils types wouldn't error here.
* Comment out line that sends bid and signed block to relay monitor.
  * This is only in Bellatrix right now, do we still want this? 

## ⛱ Motivation and Context

Since Capella, things are kind of a mixed bag; some types from go-boost-utils & some types from Attestant. I think it would be a good idea to consolidate to the Attestant types, which I feel are pretty high quality.

## 📚 References

A new release with this PR in go-boost-utils must be merged first:

* https://github.com/flashbots/go-boost-utils/pull/71

---

* [ ] `make lint`
* [ ] `make test-race`
* [ ] `go mod tidy`

### DIFF
--- a/cmd/test-cli/main.go
+++ b/cmd/test-cli/main.go
@@ -8,8 +8,11 @@ import (
 	"os"
 	"strconv"
 
+	builderApi "github.com/attestantio/go-builder-client/api"
+	apiV1Bellatrix "github.com/attestantio/go-eth2-client/api/v1/bellatrix"
+	"github.com/attestantio/go-eth2-client/spec/phase0"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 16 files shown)


metachris:
I think before moving on with this, we should remove all the bellatrix stuff from mev-boost? this will make switching the types a lot easier.

PR #498, opened 2023-04-23
add per-slot-uid to getHeader and getPayload

### LINKED ISSUE #492: Add a unique request ID to `getHeader` and `getPayload` calls.
From the relay perspective, it is quite difficult to determine when a validator for a specific slot calls `getHeader`. Because `getHeader` is a public API and the builders use it to check the value of the highest bid, there is a large amount of noise in the logs. Over the last hour, there have been 673k calls to `getHeader` on the ultra sound relay. In order to try to correlate these calls to `getPayload` requests, we can look at the IP addresses of the requests, but this is inconvenient and we would rather not handle that type of PII. 

Collecting data on when a proposer calls `getHeader` is useful because it provides insight into (1) the network latency between the validator and the relay, (2) durations for how long signing headers takes for different validators, (3) if any validators are systematically playing latency games to try to increase the amount of MEV they earn by calling `getHeader` later into the slot (See [ROP-0](https://efdn.notion.site/e57d2e4d207a43c4abdaef71456e59ae?v=3aa44b30109a40aa90f2119d60c4d075&p=385f0f6279374a90b52bf380ed76a85b&pm=s)). 

If we randomly generated a UUID for each proposer for each slot, we would be able to identify which `getHeader` calls are associated with the `getPayload` that ends up being delivered by the relay. Because `getPayload` requires a signature verification of the proposer, we can start there and then scrape our logs for calls to `getHeader` with the same UUID.

metachris:
## 📝 Summary

Add a unique ID per slot to getHeader and getPayload.

Closes #492

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -20,6 +20,7 @@ import (
 	"github.com/flashbots/go-boost-utils/types"
 	"github.com/flashbots/go-utils/httplogger"
 	"github.com/flashbots/mev-boost/config"
+	"github.com/google/uuid"
 	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
 )
@@ -47,6 +48,11 @@ type httpErrorResp struct {
 type AuctionTranscript struct {
 	Bid        *SignedBuilderBid               `json:"bid"`
 	Acceptance *types.SignedBlindedBeaconBlock `json:"acceptance"`
+}
+
+type slotUID struct {
+	slot uint64
+	uid  uuid.UUID
 }
 
 // BoostServiceOpts provides all available options for use with NewBoostService
@@ -81,8 +87,11 @@ type BoostService struct {
 	httpClientRegVal     http.Client
 	requestMaxRetries    int
 
-	bidsLock sync.Mutex
 	bids     map[bidRespKey]bidResp // keeping track of bids, to log the originating relay on withholding
+	bidsLock sync.Mutex
+
+	slotUID     *slotUID
+	slotUIDLock sync.Mutex
 }
 
 // NewBoostService created a new BoostService
@@ -104,6 +113,7 @@ func NewBoostService(opts BoostServiceOpts) (*BoostService, error) {
 		relayCheck:    opts.RelayCheck,
 		relayMinBid:   opts.RelayMinBid,
 		bids:          make(map[bidRespKey]bidResp),
+		slotUID:       &slotUID{},
 
 		builderSigningDomain: builderSigningDomain,
 		httpClientGetHeader: http.Client{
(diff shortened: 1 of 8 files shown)


server/service.go
```diff
@@ -601,10 +630,23 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
 		relays = m.relays
 	}
 
+	// Get the uid for this slot
+	uid := ""
+	m.slotUIDLock.Lock()
+	if m.slotUID.slot == uint64(payload.Message.Slot) {
+		uid = m.slotUID.uid.String()
```
ralexstokes, line 637: this works but I will note is somewhat fragile -- rather than keeping a single UUID at a time, the node could just keep a map keyed by the `getHeader` request data `(slot, pubkey, parentHash)` along w/ the UUID

there could be a background goroutine to clean up this in-memory data structure so it doesn't just grow unbounded over time

I am thinking it would be good to have more precise keying in the situation where this feature would be most important -- debugging relay behavior in an elevated forking or reorging regime

ralexstokes (commented):
generally looks good! I'd consider more precise keying that is more flexible and "harder to get wrong", esp during times where this feature is most needed

server/service.go
```diff
@@ -506,7 +533,7 @@ func (m *BoostService) processBellatrixPayload(w http.ResponseWriter, req *http.
 			log.Debug("calling getPayload")
 
 			responsePayload := new(types.GetPayloadResponse)
-			_, err := SendHTTPRequestWithRetries(requestCtx, m.httpClientGetPayload, http.MethodPost, url, ua, payload, responsePayload, m.requestMaxRetries, log)
+			_, err := SendHTTPRequestWithRetries(requestCtx, m.httpClientGetPayload, http.MethodPost, url, ua, nil, payload, responsePayload, m.requestMaxRetries, log)
```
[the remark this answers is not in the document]
metachris: yes, because this code branch is not used in mainnet or testing, and will likely be removed

server/service.go
```diff
@@ -81,8 +87,11 @@ type BoostService struct {
 	httpClientRegVal     http.Client
 	requestMaxRetries    int
 
-	bidsLock sync.Mutex
 	bids     map[bidRespKey]bidResp // keeping track of bids, to log the originating relay on withholding
+	bidsLock sync.Mutex
```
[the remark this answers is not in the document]
metachris: I generally use the pattern of the lock after the field it protects, but I wouldn't mind switching it around.

server/service.go
metachris: yeah, i guess you are right that this might be even more useful!

michaelneuder: i totally get where @ralexstokes is coming from here, but its not clear to me that adding the complexity will really make debugging much easier. we log the UUID on the validator side and we will also pick it up on the relay side and that should be plenty to get the debugging done. the slot and pubkey will be in the relay logs right by the UUID. also if there is an attacker calling getHeader with different parentHashes, then (1) we catch that on the relay end and log at the WARN level, and (2) they probably will modify their boost client to not send the UUID at all!

server/service.go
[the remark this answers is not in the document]
ralexstokes: anything goes as long as it is consistent imo :)

server/service.go
metachris: unless there's further comments, we'll merge the current implementation and extend the granularity once needed

ralexstokes (dismissed):

metachris:
Merging now, since it was re-reviewed by @ralexstokes already, and I added a commit with only a minor header key change

PR #508, opened 2023-05-13
log ms into slot when getHeader/getPayload requests start

metachris:
## 📝 Summary

Log how late into the slot the `getHeader` and `getPayload` requests start. 

This is particularly useful if proposers submit logs to debug the "too late request" response from the relay (i.e. when the relay received the request >4 sec into the slot). To be specific, this can tell whether the request started too late in the first place, or if it's the network between the proposer and relay

Moved the `GetEnv*` methods to a new `common` package, which also holds the default (mainnet) genesis time and slot time (12 sec). This removes the import of `go-utils/cli` across the package too!

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ lint:
 	staticcheck ./...
 	golangci-lint run
 
+.PHONY: lt
+lt: lint test
+
 .PHONY: fmt
 fmt:
 	gofmt -s -w .
--- a/common/common.go
+++ b/common/common.go
@@ -1,0 +1,58 @@
+package common
+
+import (
+	"math/big"
+	"os"
+	"strconv"
+
+	"github.com/flashbots/go-boost-utils/types"
+)
+
+const (
+	GenesisTimeMainnet = 1606824023
+	SlotTimeSecMainnet = 12
+)
+
+func GetEnv(key, defaultValue string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return defaultValue
+}
+
+func GetEnvInt(key string, defaultValue int) int {
+	if value, ok := os.LookupEnv(key); ok {
+		val, err := strconv.Atoi(value)
+		if err == nil {
+			return val
+		}
+	}
+	return defaultValue
+}
+
+func GetEnvFloat64(key string, defaultValue float64) float64 {
+	if value, ok := os.LookupEnv(key); ok {
+		val, err := strconv.ParseFloat(value, 64)
+		if err == nil {
+			return val
+		}
+	}
+	return defaultValue
+}
+
+// FloatEthTo256Wei converts a float (precision 10) denominated in eth to a U256Str denominated in wei
+func FloatEthTo256Wei(val float64) (*types.U256Str, error) {
+	weiU256 := new(types.U256Str)
+	ethFloat := new(big.Float)
+	weiFloat := new(big.Float)
+	weiFloatLessPrecise := new(big.Float)
+	weiInt := new(big.Int)
+
+	ethFloat.SetFloat64(val)
+	weiFloat.Mul(ethFloat, big.NewFloat(1e18))
+	weiFloatLessPrecise.SetString(weiFloat.String())
+	weiFloatLessPrecise.Int(weiInt)
+
+	err := weiU256.FromBig(weiInt)
+	return weiU256, err
+}
--- a/server/service.go
+++ b/server/service.go
@@ -319,6 +319,17 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
+	// If genesisTime and slotTimeSec are set, log how late into the slot the request started
+	if config.GenesisTime > 0 && config.SlotTimeSec > 0 {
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -44,6 +44,9 @@ lint:
 	staticcheck ./...
 	golangci-lint run
 
+.PHONY: lt
+lt: lint test
```
avalonche, line 48: the existing lint command doesn't lint tests?

common/common.go
```diff
@@ -0,0 +1,58 @@
...
+	"os"
+	"strconv"
+
+	"github.com/flashbots/go-boost-utils/types"
+)
+
+const (
+	GenesisTimeMainnet = 1606824023
+	SlotTimeSecMainnet = 12
+)
```
avalonche, line 14: I would have this grouped under types.go with all the network constants or separate into a new const file

avalonche (dismissed):

Makefile
metachris: This is just an alias that runs both lint and test

common/common.go
metachris: maybe types.go should go into the common package

metachris:
Added one more commit to log with Info level when the getPayload request starts. This will be useful when debugging bug reports like https://github.com/flashbots/mev-boost/issues/519

server/service.go
```diff
@@ -319,6 +319,17 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
+	// If genesisTime and slotTimeSec are set, log how late into the slot the request started
+	if config.GenesisTime > 0 && config.SlotTimeSec > 0 {
```
michaelneuder: cant we log without this check ? if the default is them being set to the constants, then it feels like we should always log.

server/service.go
```diff
@@ -584,6 +595,17 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
 		"parentHash": payload.Message.Body.ExecutionPayloadHeader.ParentHash.String(),
 	})
 
+	// If genesisTime and slotTimeSec are set, log how much into the slot the request started
+	if config.GenesisTime > 0 && config.SlotTimeSec > 0 {
```
michaelneuder: same comment

michaelneuder (approved):
tiny comments, but LGTM

michaelneuder (dismissed):
tiny comments, but LGTM

server/service.go
metachris: sure. this was intended as a guard against user overwriting this with weird times, but that seems overly cautious and not necessary

ralexstokes: that or we make `GetEnvUint` utils, and only allow unsigned ints

server/service.go
```diff
@@ -319,6 +319,17 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
 
+	// If genesisTime and slotTimeSec are set, log how late into the slot the request started
+	if config.GenesisTime > 0 && config.SlotTimeSec > 0 {
+		slotStartTimestamp := config.GenesisTime + (int64(_slot) * config.SlotTimeSec)
+		msIntoSlot := time.Now().UTC().UnixMilli() - (slotStartTimestamp * 1000)
+		log = log.WithField("msIntoSlot", msIntoSlot)
+		log.WithFields(logrus.Fields{
+			"genesisTime": config.GenesisTime,
+			"slotTimeSec": config.SlotTimeSec,
+		}).Infof("getHeader request start at %d milliseconds into slot %d", msIntoSlot, _slot)
```
ralexstokes: this is likely a usability thing; and it's already done so lets just keep it

but we could also just log the `time.Now().UTC().UniMilli()` and the `_slot` and do the relevant conversions in analysis later -- means less code to keep/maintain/ensure is correct here

ralexstokes (commented):
makes sense and this PR in isolation generally makes sense

do we also want to log timings on the "backend" of each request, rather than just the "frontend" like we do in this PR?

server/service.go
metachris: nah, why add a manual computation step -- for debugging it's way better to log it all together in this line!

server/service.go
metachris: it already is overwritable through GetEnvUint

metachris:
> do we also want to log timings on the "backend" of each request, rather than just the "frontend" like we do in this PR?

what do you mean with that?

ralexstokes:
right now it just timestamps when the request starts, we could also log the time when the request finishes (the "backend")

e.g. add a log on 705 about finishing to match the log on 605 about starting

metachris:
> when the request finishes (the "backend")

this is not what's typically referred to as "the backend".

I don't think we need another log there, we [already have a log](https://github.dev/flashbots/mev-boost/blob/41aafa4591acd777fc1a932e82fd7415df48db4c/server/service.go#L690) when the payload was received from the relay, and at the end of the function, the HTTP middleware also logs the complete request summary

michaelneuder (approved):
lgtm!

PR #523, opened 2023-06-05
Upgrade go-boost-utils

avalonche:
## 📝 Summary

Upgrade all the types to attestant by upgrading go-boost-utils in https://github.com/flashbots/go-boost-utils/pull/76. Tidying up types in the `common/` folder will be in a follow up PR.

## ⛱ Motivation and Context

We want to move away from go-boost-utils types entirely and move all types to attestant. This PR is part of the migration move entirely to attestant, switching the util functions to attestant types.

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service_test.go
+++ b/server/service_test.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/attestantio/go-builder-client/api"
+	apiv1 "github.com/attestantio/go-builder-client/api/v1"
 	"github.com/attestantio/go-builder-client/spec"
 	apiv1capella "github.com/attestantio/go-eth2-client/api/v1/capella"
 	consensusspec "github.com/attestantio/go-eth2-client/spec"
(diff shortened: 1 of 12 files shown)


metachris:
We'll remove the bellatrix codepaths first, which will make this update quite a bit simpler.

server/service_test.go
```diff
@@ -537,199 +539,22 @@ func TestGetHeaderBids(t *testing.T) {
 		require.Equal(t, 1, backend.relays[0].GetRequestCount(path))
 
 		// Value should be 12345 (min bid is 12345)
-		resp := new(types.GetHeaderResponse)
```
avalonche, line 540: duplicate code originally used to test capella. Now that bellatrix get header is removed, we can remove these tests

metachris (approved):
lgtm, nice work!

PR #531, opened 2023-06-16
getPayload request to all relays

### LINKED ISSUE #530: Should mev-boost to call getPayload on all relays (instead of only those which sent the bid)?
Could be useful for additional data availability -- in case relay which sent the bid can't retrieve the payload, other relays might still possess them, even if they haven't sent the bid.

Only downside I can see is that it will cause light additional load and logging on all relays, if getPayload gets called on all of them for most slots. Doesn't seem big of a downside really.

metachris:
## 📝 Summary

Send getPayload request to all relays as discussed in #530

Fixes #530

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


ralexstokes (approved):
looks good on the mev-boost side

I'm not sure how the relays will respond but that can be handled elsewhere

PR #536, opened 2023-06-22
log slot-uid

metachris:
## 📝 Summary

Instead of just sending it as part of the request, also log the slot uid with the getHeader/getPayload log messages. This should further help debugging.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -337,9 +337,9 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		m.slotUID.slot = _slot
 		m.slotUID.uid = uuid.New()
 	}
-	uid := m.slotUID.uid
+	slotUID := m.slotUID.uid
 	m.slotUIDLock.Unlock()
-	log = log.WithField("slotUid", uid)
+	log = log.WithField("slotUID", slotUID)
 
 	// Log how late into the slot the request starts
 	slotStartTimestamp := config.GenesisTime + (int64(_slot) * config.SlotTimeSec)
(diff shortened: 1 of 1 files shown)


server/service.go
```diff
@@ -348,6 +339,16 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	}
 	uid := m.slotUID.uid
 	m.slotUIDLock.Unlock()
+	log = log.WithField("slotUid", uid)
```
michaelneuder: nit: `slotUID` for consistency?

server/service.go
```diff
@@ -614,13 +615,24 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
...
+	m.slotUIDLock.Unlock()
+
 	// Prepare logger
 	ua := UserAgent(req.Header.Get("User-Agent"))
 	log = log.WithFields(logrus.Fields{
 		"ua":         ua,
 		"slot":       payload.Message.Slot,
 		"blockHash":  payload.Message.Body.ExecutionPayloadHeader.BlockHash.String(),
 		"parentHash": payload.Message.Body.ExecutionPayloadHeader.ParentHash.String(),
+		"slotUid":    uid,
```
michaelneuder: same comment as above

michaelneuder (dismissed):

server/service.go
metachris: done

server/service.go
metachris: done

michaelneuder (approved):
lg

PR #537, opened 2023-06-23
Remove process bellatrix payload

avalonche:
## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


server/service.go
```diff
@@ -221,22 +221,6 @@ func (m *BoostService) sendValidatorRegistrationsToRelayMonitors(payload []types
 	}
 }
 
-func (m *BoostService) sendAuctionTranscriptToRelayMonitors(transcript *AuctionTranscript) {
```
metachris, line 224: why remove this?

server/service.go
```diff
@@ -655,9 +525,6 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
 		log.Warn("bid found but no associated relays")
 	}
 
-	// send bid and signed block to relay monitor with capella payload
-	// go m.sendAuctionTranscriptToRelayMonitors(&AuctionTranscript{Bid: originalBid.response.Data, Acceptance: payload})
```
metachris: better keep around

metachris:
looks great on quick check! but let's keep the relay monitor stuff around for now

avalonche:
as the relay monitor code is not used, it's failing linting. instead of removing I'll just comment out the relay monitor related code for now.

metachris (approved):

Issue #529, opened 2023-06-14
Removing Bellatrix types


metachris:
As discussed in Eth R&D and other places, we plan to move forward with removing all the old Bellatrix types from mev-boost. We'll tag the latest release that contains Bellatrix support as such.

metachris:
See also https://github.com/flashbots/mev-boost-relay/issues/430

avalonche:
all bellatrix types are removed

PR #538, opened 2023-06-26
Remove wrapper types for get header response

avalonche:
## 📝 Summary

Previously wrapper types were made to process both bellatrix types from `go-boost-utils` and attestant types which includes capella types. As part of #529 this removes wrapper types as the bellatrix types from `go-boost-utils` will no longer be needed.

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


metachris:
please rebase

metachris (approved):

PR #553, opened 2023-07-28
Add Deneb Types and Flow

avalonche:
## 📝 Summary

Part of #554 

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
...
+					"requestBlobs": len(commitments),
+				}).Error("block KZG commitment length does not equal responseBlobs length")
+				return
+			}
+
+			for i, commitment := range commitments {
+				if commitment != blobs.Commitments[i] {
+					log.WithFields(logrus.Fields{
+						"requestBlobCommitment": commitment.String(),
+						"responseBlobCommiment": blobs.Commitments[i].String(),
```
jtraglia:
```suggestion
						"responseBlobCommitment": blobs.Commitments[i].String(),
```

server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
...
+						"index":                 i,
+					}).Error("requestBlobCommitment does not equal responseBlobCommiment")
```
jtraglia:
```suggestion
					}).Error("requestBlobCommitment does not equal responseBlobCommitment")
```

server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
...
+				}).Error("requestBlockHash does not equal responseBlockHash")
+				return
+			}
+
+			commitments := blindedBlock.Message.Body.BlobKZGCommitments
+			// Ensure that blobs are valid and matches the request
+			if len(commitments) != len(blobs.Blobs) || len(commitments) != len(blobs.Commitments) || len(commitments) != len(blobs.Proofs) {
+				log.WithFields(logrus.Fields{
+					"requestBlobs": len(commitments),
+				}).Error("block KZG commitment length does not equal responseBlobs length")
```
jtraglia, line 747: I think we should add `len(blobs.Commitments)`, `len(blobs.Proofs)`, and `len(blobs.Blobs)` to the `WithFields` arguments.

server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
...
+				return
+			}
+
+			payload := responsePayload.Deneb.ExecutionPayload
+			blobs := responsePayload.Deneb.BlobsBundle
+
+			// Ensure the response blockhash matches the request
+			if blindedBlock.Message.Body.ExecutionPayloadHeader.BlockHash != payload.BlockHash {
+				log.WithFields(logrus.Fields{
+					"responseBlockHash": responsePayload.Capella.BlockHash.String(),
```
jtraglia: So this is using the `Capella` field. This is wrong right? Also, I think we should log both values (the blinded block's blockHash and the payload's blockHash) so we have something to compare the wrong value to.

jtraglia (changes requested):
Here's an initial review with some things to fix. Will continue reviewing.

server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
 	m.respondOK(w, result)
 }
 
+func (m *BoostService) processDenebPayload(w http.ResponseWriter, req *http.Request, log *logrus.Entry, blindedBlock *eth2ApiV1Deneb.SignedBlindedBeaconBlock) {
+	// Get the slotUID for this slot
+	slotUID := ""
```
jtraglia: I think we should rename this variable so it doesn't shadow the type:

https://github.com/flashbots/mev-boost/blob/c91b043e382fb1e7c0ac496b10ed0f715ea99737/server/service.go#L60-L63

server/utils.go
```diff
@@ -312,3 +252,21 @@ func checkRelaySignature(bid *spec.VersionedSignedBuilderBid, domain phase0.Doma
...
+func getPayloadResponseIsEmpty(payload *builderApi.VersionedSubmitBlindedBlockResponse) bool {
+	switch payload.Version {
+	case spec.DataVersionCapella:
+		if payload.Capella == nil || payload.Capella.BlockHash == nilHash {
+			return true
+		}
+	case spec.DataVersionDeneb:
+		if payload.Deneb == nil || payload.Deneb.ExecutionPayload == nil ||
+			payload.Deneb.ExecutionPayload.BlockHash == nilHash ||
+			payload.Deneb.BlobsBundle == nil || payload.Deneb.BlobsBundle.Blobs == nil {
```
jtraglia: I don't think it's necessary to check if `Blobs` is `nil` here. It's not a pointer.

image

IIRC a nil slice will be treated like a zero-length slice.

server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
...
+		"msIntoSlot":  msIntoSlot,
+	}).Infof("submitBlindedBlock request start - %d milliseconds into slot %d", msIntoSlot, blindedBlock.Message.Slot)
+
+	// Get the bid!
+	bidKey := bidRespKey{slot: uint64(blindedBlock.Message.Slot), blockHash: blindedBlock.Message.Body.ExecutionPayloadHeader.BlockHash.String()}
+	m.bidsLock.Lock()
+	originalBid := m.bids[bidKey]
+	m.bidsLock.Unlock()
+	if originalBid.response.IsEmpty() {
+		log.Error("no bid for this getPayload payload found. was getHeader called before?")
```
jtraglia: Just to satisfy my desire to capitalize "was"...
```suggestion
		log.Error("no bid for this getPayload payload found, was getHeader called before?")
```

server/service.go
jtraglia: Just to note, had the wrong `BlockHash` been provided I think this would have crashed mev-boost.

server/service.go
```diff
@@ -51,8 +53,8 @@ type httpErrorResp struct {
 
 // AuctionTranscript is the bid and blinded block received from the relay send to the relay monitor
 type AuctionTranscript struct {
-	Bid        *spec.VersionedSignedBuilderBid          // TODO: proper json marshalling and unmashalling
-	Acceptance *apiv1bellatrix.SignedBlindedBeaconBlock `json:"acceptance"`
+	Bid        *builderSpec.VersionedSignedBuilderBid       // TODO: proper json marshalling and unmashalling
```
jtraglia:
```suggestion
	Bid        *builderSpec.VersionedSignedBuilderBid       // TODO: proper json marshalling and unmarshalling
```

metachris:
applied suggestions from @jtraglia and rebased

server/service.go
```diff
@@ -646,6 +651,143 @@ func (m *BoostService) processCapellaPayload(w http.ResponseWriter, req *http.Re
 	m.respondOK(w, result)
 }
 
+func (m *BoostService) processDenebPayload(w http.ResponseWriter, req *http.Request, log *logrus.Entry, blindedBlock *eth2ApiV1Deneb.SignedBlindedBeaconBlock) {
```
michaelneuder, line 654: would it be worth making a generic processPayload function that works with both deneb and capella blocks? just in terms of reducing code copying for ease of review? maybe not if we plan on stripping out the capella payload processing function immediately after the fork. just a thought as its hard to tell what logic is new with deneb and which is just copied from capella.

server/utils_test.go
```diff
@@ -97,15 +101,100 @@ func TestWeiBigIntToEthBigFloat(t *testing.T) {
 	require.Equal(t, "0.000000000000000000", f.Text('f', 18))
 }
 
-func TestCapellaComputeBlockHash(t *testing.T) {
-	jsonFile, err := os.Open("../testdata/zhejiang-execution-payload-capella.json")
-	require.NoError(t, err)
-	defer jsonFile.Close()
+func TestGetPayloadResponseIsEmpty(t *testing.T) {
```
michaelneuder, line 104: could this be written as a table driven test to improve readability ?

server/service.go
jtraglia: I think we should strip Capella shortly afterwards.

avalonche: I've thought about this approach but there are issues as the process payload function accepts different types. We would need to have a custom interface to extract the common fields and process them which leads to more potential bugs.w/ versioning. There is code duplication but less prone to issues with version handling.

metachris: good point @avalonche, agree. this is a space with high potential for critical bugs, better be safe with some duplication

server/utils_test.go
avalonche: will do as follow up PR

server/service.go
avalonche: will address as follow up

jtraglia (approved):
LGTM 🥇

metachris (approved):

PR #590, opened 2023-11-23
fix: holesky genesis timestamp

aimxhaisse:
## 📝 Summary

The current Holesky genesis timestamp used in MEV boost is not correct.

## ⛱ Motivation and Context

This prevents usage of mev-boost under the Holesky network.

## 📚 References

Genesis can be found here: https://holesky.beaconcha.in/slot/0

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


metachris (approved):

metachris:
Confirmed

PR #635, opened 2024-03-19
Set defaultRelayMinBidEth to 0.05 ETH

timbeiko:
## 📝 Summary

Defaults are powerful, and setting a non-zero `defaultRelayMinBid` can help increase censorship resistance. 

## ⛱ Motivation, Context and References

`defaultRelayMinBid` is a valuable feature to improve Ethereum's censorship resistance, but is currently off by default. 

In Nov 2022, a 0.05 ETH `MinBid` was shown to result in validators proposing ~50% of blocks locally while keeping 80% of total fees ([source](https://twitter.com/VitalikButerin/status/1595793106926014466)). With a majority of builders currently censoring transactions ([source](https://censorship.pics/)), setting a non-zero default for this value can help increase the overall censorship resistance of the network. 

While anyone can opt-out of this change trivially by setting the flag themselves, given mev-boost's broad adoption, many users are likely to stick to the defaults. 

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -35,7 +35,7 @@ var (
 	defaultLogLevel          = common.GetEnv("LOG_LEVEL", "info")
 	defaultListenAddr        = common.GetEnv("BOOST_LISTEN_ADDR", "localhost:18550")
 	defaultRelayCheck        = os.Getenv("RELAY_STARTUP_CHECK") != ""
-	defaultRelayMinBidEth    = common.GetEnvFloat64("MIN_BID_ETH", 0)
+	defaultRelayMinBidEth    = common.GetEnvFloat64("MIN_BID_ETH", 0.05)
 	defaultDisableLogVersion = os.Getenv("DISABLE_LOG_VERSION") == "1" // disables adding the version to every log entry
 	defaultDebug             = os.Getenv("DEBUG") != ""
 	defaultLogServiceTag     = os.Getenv("LOG_SERVICE_TAG")


metachris:
50% local blocks seems quite destructive for the PBS and builder market ("would throw quite a bit of chaos into the equation for searchers and order-flow providers").

jtraglia:
Personally, I'm not in favor of having a default min bid. I much prefer the solution that clients have already implemented, where the local payload will be used unless the builder's payload is 10% more valuable. That is a better solution which shouldn't require changes in the future, if for example the average payload changes.

* If in 5 years the price of ETH is $10k (or higher), is the average payload value expected to be the same? If the average goes down so that its fiat value remains about the same, this will cause an increasing number of validators to produce blocks locally. This is important if we're truly targeting 50% for local payload production.
* How will this affect testnets/devnets? The average payload value is much lower there. Will we (1) require nodes to disable the min-bid feature or (2) artificially produce payloads with values greater than 0.05 ETH to test things?

dataalways:
I'm supportive of the change, but I think it needs stronger empirical justification. 

Looking at data from proposers who already use `min-bid`, if the entire network adopted the change (`min-bid`= 0.05) proposers would lose on the order of $80 million per year in rewards. I also don't know how much I really believe in this defaults are sticky idea, the main issue is that only one of the top 20 staking entities by size (Upbit: 15th biggest) uses the parameter, are we really to believe that the top staking pools and CEXs just haven't bothered to change the default? It seems more likely to me that they just haven't been presented with proper cost-benefit analysis. It seems to me like changing the default will mostly affect solo stakers.

I would also like to see deeper discussion of the economics of client changes. It might sound like they're doing roughly equivalent things, but the distribution of blocks that they catch are vary different. The best result is probably some combination of the two methods (or looking at difference in values rather than just scaling local block values up by 10%).

Evan-Kim2028:
Agree with @dataalways there should be a stronger empirical justification than a vitalik post from November 2022, with <3 months worth of beacon chain data.
 
For the other source, although the "majority of builders" are censoring transactions, this [other dashboard](https://www.mevwatch.info/) shows that there are more non-ofac compliant blocks built than ofac compliant bllocks.

timbeiko:
Appreciate all the feedback here — will close as this clearly hasn't consensus to move forward!

PR #639, opened 2024-04-02
common: simplify floatToEth

MariusVanDerWijden:
Simplify the `FloatEthTo256Wei` calculation. The trick with converting to string is still kinda ugly imo


jtraglia (approved):
LGTM, thanks! This looks marginally better.

metachris (approved):

PR #641, opened 2024-04-02
cli: refactor sanitizeMinBid

MariusVanDerWijden:
Refactor sanitizeMinBid

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -179,6 +179,9 @@ func Main() {
 	if err != nil {
 		log.WithError(err).Fatal("Failed sanitizing min bid")
 	}
+	if relayMinBidWei.BigInt().Sign() > 0 {
+		log.Infof("Min bid set to %v wei", relayMinBidWei)
+	}
 
 	opts := server.BoostServiceOpts{
 		Log:                      log,
@@ -252,9 +255,6 @@ func sanitizeMinBid(minBid float64) (*types.U256Str, error) {
 	}
 	if minBid > 1000000.0 {
 		return nil, errors.New("minimum bid is too large, please ensure min-bid is denominated in Ethers")
-	}
-	if minBid > 0.0 {
-		log.Infof("Minimum bid: %v eth", minBid)
 	}
 	return common.FloatEthTo256Wei(minBid)
 }
--- a/cli/main.go
+++ b/cli/main.go
@@ -177,7 +177,7 @@ func Main() {
 
 	relayMinBidWei, err := sanitizeMinBid(*relayMinBidEth)
 	if err != nil {
-		log.WithError(err).Fatal("failed sanitizing min bid")
+		log.WithError(err).Fatal("Failed sanitizing min bid")
 	}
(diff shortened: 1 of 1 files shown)


cli/main.go
```diff
@@ -256,3 +245,16 @@ func setupLogging() error {
 	log.Debug("debug logging enabled")
 	return nil
 }
+
+func sanitizeMinBid(minBid float64) (*types.U256Str, error) {
+	if minBid < 0.0 {
+		log.Fatal("Please specify a non-negative minimum bid")
```
avalonche: can return error instead of log.Fatal and leave the error check to the calling function

MariusVanDerWijden: yep

cli/main.go
```diff
@@ -256,3 +245,16 @@ func setupLogging() error {
...
+
+func sanitizeMinBid(minBid float64) (*types.U256Str, error) {
+	if minBid < 0.0 {
+		return nil, errors.New("please specify a non-negative minimum bid")
+	}
+	if minBid > 1000000.0 {
+		return nil, errors.New("minimum bid is too large, please ensure min-bid is denominated in Ethers")
+	}
+	if minBid > 0.0 {
+		log.Infof("Minimum bid: %v eth", minBid)
```
avalonche: nit: should log after error checking and should not be in `sanitizeMinBid`

avalonche (dismissed):

MariusVanDerWijden:
The linter fails, because it complains about dynamic errors and enforces static errors. I kinda disagree with the linter. Declaring static errors makes everything more spaghetti and this method is not used from a context where the type checking you get with static errors is useful. I am going to declare them as static anyway, to make the linter happy, but I think it makes the code worse and you should consider disabling the `err113` check

cli/main.go
```diff
@@ -174,21 +175,12 @@ func Main() {
...
-	}
-
-	relayMinBidWei, err := common.FloatEthTo256Wei(*relayMinBidEth)
+	relayMinBidWei, err := sanitizeMinBid(*relayMinBidEth)
 	if err != nil {
-		log.WithError(err).Fatal("failed converting min bid")
+		log.WithError(err).Fatal("Failed sanitizing min bid")
+	}
+	if relayMinBidWei.BigInt().Sign() > 0 {
+		log.Infof("Min bid set to %v wei", relayMinBidWei)
```
jtraglia: This changes the log from Eth to Wei. Maybe we should log both.

```go
log.Infof("Min bid set to %v eth (%v wei)", relayMinBidEth, relayMinBidWei)
```

jtraglia (dismissed):
LGTM, just one suggestion around min bid log.

jtraglia (approved):

metachris (approved):
Looks great, thanks!

About the error linter -- you making a strong case definitely means a lot to me and I'd like to consider it. Do you feel strongly enough to create an issue on this repo, to continue the conversation there?

PR #643, opened 2024-04-11
server: refactor handleGetHeader

MariusVanDerWijden:
## 📝 Summary

This pulls out the verification logic into verifyBid which makes it easier to unit test in the future. It also changes the logic for parallelizing the requests to the builder from using a shared mutex to computing the requests in parallel and assembling the best response after the fact on the main thread

## ⛱ Motivation and Context

This PR makes verifyBid easier to unit test as it does not depend on setting up multiple relays etc.
This PR also removes some very un-standard logic of using a shared mutex to compute the best bid.

I am not 100% sure about the performance implications of this PR, but I think it should be faster in my opinion, since we do not need to wait for the locking of the mutex

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -343,12 +343,15 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		HeaderKeySlotUID: slotUID.String(),
 	}
 
-	// Prepare relay responses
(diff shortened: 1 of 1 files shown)


jtraglia:
Going to close this. No longer relevant.

PR #644, opened 2024-04-11
server: refactor process*Payload functions

MariusVanDerWijden:
This PR changes the way the process functions work. Since it is safe to call `cancel()` multiple times, and we only need one valid payload from the relays, I think its safe to stick the results in a channel and only query one of them


server/service.go
```diff
@@ -737,26 +727,19 @@ func (m *BoostService) processDenebPayload(w http.ResponseWriter, req *http.Requ
...
-
-			if requestCtx.Err() != nil { // request has been cancelled (or deadline exceeded)
-				return
-			}
-
-			// Received successful response. Now cancel other requests and return immediately
 			requestCtxCancel()
-			*result = *responsePayload
+			resultCh <- responsePayload
 			log.Info("received payload from relay")
```
jtraglia: I think there could be value in adding back a cancellation check here. There's also a check at the beginning of the function, but this one (closer to the end) is most likely to catch it, right? This would help prevent multiple, relatively verbose "received payload from relay" messages.

server/service.go
```diff
@@ -737,26 +727,19 @@ func (m *BoostService) processDenebPayload(w http.ResponseWriter, req *http.Requ
...
 		}(relay)
 	}
 
 	// Wait for all requests to complete...
 	wg.Wait()
+	close(resultCh)
+	result := <-resultCh
```
jtraglia: This is a strange pattern, but it should work as intended. We still need the wait group unfortunately. Normally, reading from a channel is a blocking operation, but if the channel is closed it's not blocking. It would be cool if we could figure out a way of handling this better, where we don't necessarily have to wait for each request to finish. As in, remove the wait group and proceed as soon as we receive the first request.

jtraglia (dismissed):
I like the spirit of this PR. Removing the write lock is nice. Channels are useful. I think it would be good to add back the deleted cancellation check, see my comment for more details.

server/service.go
MariusVanDerWijden: I added back a check that should get rid of multiple received payload from relay messages, but its kinda ugly. I kinda prefer the multiple log lines, ptal

server/service.go
MariusVanDerWijden: I removed the wg and changed the pattern so that we only need one request to finish / or the timeout to trigger

MariusVanDerWijden:
Somethings wrong... am investigating

MariusVanDerWijden:
Seems like the issue was the same as in https://github.com/flashbots/mev-boost/pull/650
Attestantio was broken

metachris (dismissed):
This is a nice refactor and simplification, thanks 👍

metachris (approved):

PR #647, opened 2024-04-24
Add request start time metrics to request header field

bhakiyakalimuthu:
## 📝 Summary
This PR introduces the ability to add the request start time in Unix milliseconds and the start time into the slot to the request headers. This enhancement facilitates the measurement of latency between the validator and the relay.

## ⛱ Motivation and Context
By adding these time metrics into the headers, we can better analyze the causes of request delays, determining whether they are due to network latency or performance issues with the validator node.

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/service.go
+++ b/server/service.go
@@ -257,6 +257,11 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 		"ua":               ua,
 	})
 
+	// Add request headers
+	headers := map[string]string{
+		HeaderStartTimeMsUnix: fmt.Sprintf("%d", time.Now().UTC().UnixMilli()),
+	}
+
 	relayRespCh := make(chan error, len(m.relays))
 
 	for _, relay := range m.relays {
@@ -264,7 +269,7 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
 			url := relay.GetURI(params.PathRegisterValidator)
 			log := log.WithField("url", url)
 
-			_, err := SendHTTPRequest(context.Background(), m.httpClientRegVal, http.MethodPost, url, ua, nil, payload, nil)
+			_, err := SendHTTPRequest(context.Background(), m.httpClientRegVal, http.MethodPost, url, ua, headers, payload, nil)
 			relayRespCh <- err
 			if err != nil {
 				log.WithError(err).Warn("error calling registerValidator on relay")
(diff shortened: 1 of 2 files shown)


server/service.go
```diff
@@ -338,16 +343,15 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		"msIntoSlot":  msIntoSlot,
 	}).Infof("getHeader request start - %d milliseconds into slot %d", msIntoSlot, _slot)
 
-	// Add request headers
```
bhakiyakalimuthu, line 341: FYI-I had to remove 3 commented line to make linter happy. Otherwise `handleGetHeader` method having below linter issue.
` Function name: handleGetHeader, Cyclomatic Complexity: 24, Halstead Volume: 6038.34, Maintainability Index: 19`

server/service.go
```diff
@@ -337,16 +342,15 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		"slotTimeSec": config.SlotTimeSec,
 		"msIntoSlot":  msIntoSlot,
 	}).Infof("getHeader request start - %d milliseconds into slot %d", msIntoSlot, _slot)
-
 	// Add request headers
 	headers := map[string]string{
-		HeaderKeySlotUID: slotUID.String(),
+		HeaderKeySlotUID:          slotUID.String(),
+		HeaderStartTimeMsIntoSlot: fmt.Sprintf("%d", msIntoSlot),
```
avalonche: why do we need to add msIntoSlot in the header? `HeaderStartTimeMsUnix` should contain the same information

bhakiyakalimuthu: Ok. I have removed `HeaderStartTimeMsUnix` from getHeader adn getPayload.

avalonche: I think you should remove `HeaderStartTimeMsIntoSlot` and have `HeaderStartTimeMsUnix` consistently across the endpoints

bhakiyakalimuthu: ok.Agreed. Removed `HeaderStartTimeMsUnix` and made consistent across endpoint

avalonche (approved):

jtraglia (approved):
This seems reasonable 👍 good idea.

Can we add back the deleted blank lines? This is just a nit.

bhakiyakalimuthu:
> This seems reasonable 👍 good idea.
> 
> Can we add back the deleted blank lines? This is just a nit.

Those lines were deleted for a reason. Linter was throwing error as below if the lines were not removed. 
`server/service.go:295:1: Function name: handleGetHeader, Cyclomatic Complexity: 24, Halstead Volume: 5978.93, Maintainability Index: 19 (maintidx)
func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request) {
^
make: *** [lint] Error 1
`

avalonche:
interesting that removing blank lines makes the linter think there is less complexity. the function probably needs refactoring

bhakiyakalimuthu:
> interesting that removing blank lines makes the linter think there is less complexity. the function probably needs refactoring

Hi @avalonche  Is there anything else we need to address, or are we ready to merge this PR?

PR #651, opened 2024-04-29
Add support for Electra

jtraglia:
## 📝 Summary

This PR adds support for Electra payloads.

## ⛱ Motivation and Context

* https://github.com/ethereum/consensus-specs/tree/dev/specs/electra

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


avalonche (dismissed):
LGTM, are tests for this going to be a follow up PR?

jtraglia:
> LGTM, are tests for this going to be a follow up PR?

Yes, I'm planning to add tests later. Please hold off on merging this for a little bit. We're working on getting the whole Electra builder stack working in Kurtosis this week. Thank you for reviewing though 😄

I'm fairly confident in these mev-boost changes, but mev-boost-relay will need some more work.

metachris (dismissed):

metachris (commented):
types are out of date by now

MariusVanDerWijden:
Waiting for https://github.com/attestantio/go-eth2-client/pull/163
Once thats merged, we can update this

metachris (approved):

go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/flashbots/mev-boost
 
-go 1.21
+go 1.22
 
 require (
-	github.com/ethereum/go-ethereum v1.14.7
-	github.com/flashbots/go-boost-utils v1.8.1
+	github.com/ethereum/go-ethereum v1.14.9
```
ryanschneider, line 6: minor: I'd suggest bumping to 1.14.13 while we're updating.

ryanschneider (approved):
lgtm!

server/service_test.go
```diff
@@ -125,6 +126,41 @@ func blindedBlockContentsToPayloadDeneb(signedBlindedBlockContents *eth2ApiV1Den
 	}
 }
 
+func blindedBlockContentsToPayloadElectra(signedBlindedBlockContents *eth2ApiV1Electra.SignedBlindedBeaconBlock) *builderApiDeneb.ExecutionPayloadAndBlobsBundle {
```
MariusVanDerWijden, line 129: shouldn't this return `*builderApiElectra` instead of `builderApiDeneb`?

testdata/signed-blinded-beacon-block-electra.json
```diff
@@ -0,0 +1,72 @@
...
+        "gas_limit": "30000000",
+        "gas_used": "84000",
+        "timestamp": "1732296378",
+        "extra_data": "0x4e65746865726d696e64",
+        "base_fee_per_gas": "7",
+        "block_hash": "0xb65b77e52407ff25f7fcd3f455c991ae67be1bc10cb7ec992a659698cf88870f",
+        "transactions_root": "0xb1fcd304d8ba402be6e76346395ea7641fbb4c83663b697a0e88ab115d859d3f",
+        "withdrawals_root": "0x792930bbd5baac43bcc798ee49aa8185ef76bb3b44ba62b91d86ae569e4bb535",
+        "blob_gas_used": "0",
+        "excess_blob_gas": "0"
```
MariusVanDerWijden, line 60: Isn't this missing `ParentBeaconRoot` and `RequestHash` ?

MariusVanDerWijden (commented):
Generally LGTM, two thinks that needs changing, otherwise I think the test would be moot. Also shouldn't this be caught at some point? Or is mev-boost not checking the header at all

server/service_test.go
jtraglia: Nah because there is no new `ExecutionPayloadAndBlobsBundle` in Electra.

server/service.go
```diff
@@ -631,6 +639,152 @@ func (m *BoostService) processDenebPayload(w http.ResponseWriter, req *http.Requ
 	m.respondOK(w, result)
 }
 
+func (m *BoostService) processElectraPayload(w http.ResponseWriter, req *http.Request, log *logrus.Entry, blindedBlock *eth2ApiV1Electra.SignedBlindedBeaconBlock) {
+	// Get the currentSlotUID for this slot
+	currentSlotUID := ""
+	m.slotUIDLock.Lock()
```
MariusVanDerWijden, line 645: I think we can get away with an RLock here

MariusVanDerWijden: Or make SlotUID an atomic

testdata/signed-blinded-beacon-block-electra.json
jtraglia: Hmm I don't think so. I believe you're thinking of [`NewPayloadRequest`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-newpayloadrequest).

server/service.go
MariusVanDerWijden: I will do this in a follow-up

Issue #640, opened 2024-04-02
Adding a cli library


MariusVanDerWijden:
Would you be open to move the cli/flags handling from the basic-builtin handling to some library?

ralexstokes:
I'd support this, esp for the right library -- is there one you have in mind?

I assume the intent is to simplify or harden the current flags infrastructure?

MariusVanDerWijden:
Yes exactly, I would use `https://github.com/urfave/cli` which is the one we use in go-ethereum.
Theres a lot of dereferencing of flags going on which feels a bit unsafe

Champ10q:
> Yes exactly, I would use `https://github.com/urfave/cli` which is the one we use in go-ethereum.
> Theres a lot of dereferencing of flags going on which feels a bit unsafe

Champ10q:
I can implement this and will use urfave/cli (matches go-ethereum).

Champ10q:
> Would you be open to move the cli/flags handling from the basic-builtin handling to some library?

Champ10q:
Yes> Would you be open to move the cli/flags handling from the basic-builtin handling to some library?

stlpruitt314:
Yes

PR #660, opened 2024-07-13
cmd/mev-boost: add urfave as cli framework

MariusVanDerWijden:
## 📝 Summary

This PR adds urfave/cli as a cli framework and reworks some of the logic in the main package.
It also gets rid of the weird redirections into the cli package

## ⛱ Motivation and Context

Cleanup, adds a cli framework which makes adding flags easier in the future


MariusVanDerWijden:
closes https://github.com/flashbots/mev-boost/issues/640

metachris:
Cool, thanks! A few quick notes
- The "weird indirection" was to allow `go install github.com/flashbots/mev-boost` (it won't work without a root main.go file i think)
- I like urfave a lot, and the cli flags have grown significantly and unpleasantly in mev-boost. But is the switch to urfave backwards-compatible, or would it break? This seems to be a major concern, if the upgrade would force all proposers to update their cli flags.

Makefile
```diff
@@ -24,7 +24,7 @@ v:
 .PHONY: build
 build:
 	@go version
-	CGO_ENABLED=0 go build $(GO_BUILD_FLAGS) -o mev-boost
+	CGO_ENABLED=0 go build $(GO_BUILD_FLAGS) -o mev-boost ./cmd/mev-boost
```
jtraglia, line 27: What's the rationale for this change?

jtraglia (dismissed):

metachris:
@MariusVanDerWijden can you please add the root `main.go` file back, so `go install github.com/flashbots/mev-boost` works?

Makefile
metachris: removing the root `main.go` file (i think we should keep that file and not change the build)

metachris (approved):

PR #703, opened 2024-11-22
server: refactor processElectraPayload

0xTylerHolmes:
## 📝 Summary

This PR refactors the processElectraPayload method to use a chan and have a timeout, this makes it more inline with how the deneb payload handler was implemented. It also adds a processElectraPayload test.

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


jtraglia:
The linter error is not your fault. Will fix in a separate PR.

jtraglia (approved):

PR #704, opened 2024-11-22
modify processElectraPayload to use same request headers as deneb

0xTylerHolmes:
## 📝 Summary

modify the request headers in processElectraPayload to include the request start time like the previous implementation

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


jtraglia (approved):
Good catch. Thank you!

jtraglia:
The linter error is not your fault. Will fix in a separate PR.

PR #710, opened 2024-12-20
Fix versionFlag to output to stdout

### LINKED ISSUE #47: mev-boost v1.7+ no longer outputs version to stdout
When using ethpillar to update mev-boost, the current version no longer is displayed.

coincashew:
## 📝 Summary

Rollsback a change which affected the versionFlag output
```mev-boost --version```
outputs to stdout once again.

Version output is currently outputting to stderr.

## ⛱ Motivation and Context

Fixes https://github.com/coincashew/EthPillar/issues/47

**Before**
```

```

**After**
```
mev-boost v1.8.1-6-gc574ad2-dev
```

### CODE REVIEW

File: cli/main.go [hunk 1]
```diff
@@ -55,7 +55,7 @@ func Main() {
 func start(_ context.Context, cmd *cli.Command) error {
 	// Only print the version if the flag is set
 	if cmd.IsSet(versionFlag.Name) {
-		log.Infof("mev-boost %s\n", config.Version)
+		fmt.Printf("mev-boost %s\n", config.Version) //nolint
```
metachris: probably better to do the `setupLogging` step first, which sets it to stdout, and keep using `log.Infof`.

jtraglia: I think using `fmt.Printf` here is the right thing actually. Otherwise, the output is difficult to parse:

```
$ ./mev-boost --version
INFO[2025-01-31T08:45:58.966-06:00] starting mev-boost                            version=3c51355-dev
INFO[2025-01-31T08:45:58.966-06:00] mev-boost 3c51355-dev                         version=3c51355-dev
```

And this would break EthPillar's regex here:

```
MB=$(if [[ -f /etc/systemd/system/mevboost.service ]]; then printf "Mev-boost: $(mev-boost --version | sed 's/.*\s\([0-9]*\.[0-9]*\).*/\1/')"; else printf "Mev-boost: Not Installed"; fi)
```

jtraglia: I've updated the PR to do this instead:

```go
fmt.Fprintf(cmd.Writer, "mev-boost %s\n", config.Version)
```

I believe this is most appropriate for CLI apps. Doesn't require `//nolint` either.

jtraglia (approved):

jtraglia:
Thanks @coincashew! I'm going to merge this now.

PR #715, opened 2025-01-30
Minor cleanup

MariusVanDerWijden:
I think we should split the functionality for handling the http stuff from the actual functionality of the code

### DIFF
--- a/server/functionality.go
+++ b/server/functionality.go
@@ -1,0 +1,176 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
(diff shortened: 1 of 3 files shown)


MariusVanDerWijden:
I just realized that I already tried something similar here: https://github.com/flashbots/mev-boost/pull/643

jtraglia (approved):
LGTM, thanks! I'm not sure I understand the rationale for the `bidKey` changes, but it seems fine to me. Using a structure there was explicit, which I liked. And now `fmt.Sprintf` is called inside of `m.bidsLock.Lock()` instead of before it; but this since this operation is so fast, it shouldn't really matter. The other changes (splitting functionality into another file, changes to errors, etc) are great.

PR #726, opened 2025-02-02
server: refactor TestGetPayloadFork tests

MariusVanDerWijden:
This PR refactors the TestGetPayload{Bellatrix, ...} tests


server/service_test.go
```diff
@@ -838,142 +727,182 @@ func TestEmptyTxRoot(t *testing.T) {
...
-			jsonFile, err := os.Open(fn)
+	return nil
+}
+
+func denebHeader(header *deneb.ExecutionPayloadHeader, kzgCommitments []deneb.KZGCommitment) *builderApiDeneb.ExecutionPayloadAndBlobsBundle {
+	numBlobs := len(kzgCommitments)
+	commitments := make([]deneb.KZGCommitment, numBlobs)
+	copy(commitments, kzgCommitments)
+	proofs := make([]deneb.KZGProof, numBlobs)
+	blobs := make([]deneb.Blob, numBlobs)
```
jtraglia, line 799: This feels wrong. Where are proofs and blobs populated?

server/service_test.go
```diff
@@ -838,142 +727,182 @@ func TestEmptyTxRoot(t *testing.T) {
 	require.Equal(t, "0x7ffe241ea60187fdb0187bfa22de35d1f9bed7ab061d9401fd47e34a54fbede1", txRootHex)
 }
 
-func TestGetPayloadWithTestdata(t *testing.T) {
-	path := "/eth/v1/builder/blinded_blocks"
-	testPayloadsFiles := []string{
-		"../testdata/signed-blinded-beacon-block-capella.json",
+func blindBlockToBlockResponse[P Payload](signedBlock P) *builderApi.VersionedSubmitBlindedBlockResponse {
```
jtraglia: Let's rename "blind" to "blinded"

server/service_test.go
MariusVanDerWijden: You're right, seems like it was borked before already

MariusVanDerWijden: Ah nvm I get it now, the proofs are not part of the blindedBlock, only the blindedBlockResponse, so we can't get them from the blindedBlock itself. Same with the withdrawals and transactions. blindedBlock is kindof a subset of blindedBlockResponse

jtraglia (approved):
LGTM, let's look into populating proofs & blobs later.

PR #729, opened 2025-02-03
Split functionality into three files

jtraglia:
## 📝 Summary

I think "functionality" is a bit too vague. Splitting things up makes sense IMO:

* `register_validator.go` -- implements `registerValidator`
* `get_header.go` -- implements `getHeader`
* `get_payload.go` -- implements `getPayload`

Also does some intermediate refactoring, particularly in `getPayload`.

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


server/service.go
```diff
@@ -350,7 +329,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 // respondPayload responds to the proposer with the payload
 func (m *BoostService) respondPayload(w http.ResponseWriter, log *logrus.Entry, result *builderApi.VersionedSubmitBlindedBlockResponse, originalBid bidResp) {
 	// If no payload has been received from relay, log loudly about withholding!
-	if result == nil || getPayloadResponseIsEmpty(result) {
+	if result == nil {
```
jtraglia, line 332: This is a notable change. I've removed `getPayloadResponseIsEmpty` (only exists in `verifyPayload` now). I do not believe it's necessary to check this here but I could be wrong. If the response is empty, `verifyPayload` will return a nil result and this will respond with an error.

MariusVanDerWijden: I think this prevents us from accessing nil fields. I don't know if those are set during the unmarshalling. If the are, I think we can remove this check

jtraglia:
I'm going to close this & break it up into a few PRs.

PR #733, opened 2025-02-05
Forward validator registrations without decoding

jtraglia:
## 📝 Summary

* Move functionality to new `register_validator.go` file.
* Forward validator registrations to each relay without decoding the message.
  * If the node sends validator registrations in SSZ, we inherit that.
* Add a `wrapUserAgent` method.

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/register_validator.go
+++ b/server/register_validator.go
@@ -1,0 +1,99 @@
+package server
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/flashbots/mev-boost/server/params"
+	"github.com/flashbots/mev-boost/server/types"
+	"github.com/sirupsen/logrus"
+)
+
+func (m *BoostService) registerValidator(log *logrus.Entry, regBytes []byte, header http.Header) error {
+	respErrCh := make(chan error, len(m.relays))
+
+	// Forward request to each relay
+	for _, relay := range m.relays {
+		go func(relay types.RelayEntry) {
+			// Get the URL for this relay
+			requestURL := relay.GetURI(params.PathRegisterValidator)
+			log := log.WithField("url", requestURL)
+
+			// Build the new request
+			req, err := http.NewRequestWithContext(context.Background(), http.MethodPost, requestURL.String(), bytes.NewReader(regBytes))
+			if err != nil {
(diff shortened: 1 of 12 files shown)


server/register_validator.go
```diff
@@ -0,0 +1,100 @@
...
+			// validator registrations to the other relays.
+			return nil
+		}
+	}
+
+	// None of the relays responded OK
+	return errNoSuccessfulRelayResponse
+}
+
+func (m *BoostService) sendValidatorRegistrationsToRelayMonitors(log *logrus.Entry, regBytes []byte, header http.Header) {
```
ralexstokes, line 71: i think the relay monitor code paths have been deprecated; and if not, likely should be

ralexstokes (dismissed):
the code generally looks good. some notes:

- there are no tests for this AFACIT, we should test somehow before putting into a release
- there is a change in behavior here where mev-boost no longer verifies the validator registrations deserialize correctly; it should be fine (the relay will still do this, etc) but is somewhat of a notable change
- im not going all of the go APIs with a fine-toothed comb, so could be missing something with error handling, request processing, etc

server/register_validator.go
jtraglia: Yes. I would like to remove this in a separate PR.

jtraglia:
> there are no tests for this AFACIT, we should test somehow before putting into a release

Agreed. I've pushed some basic tests to ensure the forwarding works as expected.

> there is a change in behavior here where mev-boost no longer verifies the validator registrations deserialize correctly; it should be fine (the relay will still do this, etc) but is somewhat of a notable change

Also agreed. But I feel that this is fine. The builder/relay should check this. Keep in mind it's possible for clients to connect directly to a single builder/relay and skip the mev-boost middleman. Meaning they need to handle bad inputs too.

server/register_validator.go
metachris: any change we could remove the monitor first? would simplify things a bit

metachris (approved):
nice code and docs/comments!

server/register_validator.go
jtraglia: Will remove afterwards 👍

PR #734, opened 2025-02-05
Add getHeader SSZ support

jtraglia:
## 📝 Summary

This PR adds SSZ support for `getHeader` endpoint. Please see:

* https://github.com/ethereum/builder-specs/blob/main/apis/builder/header.yaml

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


0xTylerHolmes:
I think they are fine, but two things that caught my attention: 
1. Reusing a request object (it is fine if body == nil)
2. Double close scenarios (io.Closer specifies that calls to Close after first call is considered undefined; though I think it is fine here)

First thought that addresses both of these and simplifies things a bit is if we define a doRequest function inside `get_header` that we can use to handle both the SSZ and JSON request attempts:

```go annotate 
doRequest := func(accept string) (*http.Response, error) {
                req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, url, nil)
                if err != nil {
                    return nil, err
                }
                // <insert request prep>
                req.Header.Set("Accept", accept)
                return m.httpClientGetHeader.Do(req)
            }
```

Then we can do something like: 
```go annotate 
            // ssz attempt
            resp, err := doRequest("application/octet-stream")
            if err != nil {
                …
                return
            }
            // json if relay does not support ssz
            if resp.StatusCode == http.StatusNotAcceptable {
                resp.Body.Close()
                logRelay.Debug("SSZ not accepted, trying JSON")
                resp, err = doRequest("application/json")
                if err != nil {
                    …
		    return
                }
            }
            defer resp.Body.Close()
            // continue rest of execution
            ...
```
edit: `get_header` should probably use a context that we pass to the NewRequestWithContext so we can handle timeouts.

server/get_header.go
```diff
@@ -71,18 +76,90 @@ func (m *BoostService) getHeader(log *logrus.Entry, ua UserAgent, slot phase0.Sl
...
+			case "application/octet-stream":
+				log.Debug("requesting header in SSZ")
+				resp, err = doRequest("application/octet-stream")
+				if err != nil {
+					log.WithError(err).Warn("error calling getHeader on relay")
+					return
+				}
+
+				// Check if the relay supports SSZ requests
+				if resp.StatusCode != http.StatusNotAcceptable {
```
nflaig: While this approach works I am not sure about the feasibility in practice due to time constraints, I think it would be best to use q-weighted Accept headers as [noted here](https://github.com/ethereum/builder-specs/pull/104/files#diff-9632ee7eb2afc1f79cb5a37313d0688a6518ffaf8315da0fa14761105d1faf1fR12). To handle the response body, you can look at the Content-Type header sent by server.

jtraglia: Oh interesting. So this can/should be done in a single request. I'll just forward whatever the client sends & handle the response with Content-Type like you suggest.

jtraglia: @nflaig if you have time to review this again, I would really appreciate it 🙏

server/get_header.go
```diff
@@ -71,18 +76,60 @@ func (m *BoostService) getHeader(log *logrus.Entry, ua UserAgent, slot phase0.Sl
...
+
+			// Get the resp body content
+			respBytes, err := io.ReadAll(resp.Body)
+			if err != nil {
+				log.WithError(err).Warn("error reading response body")
+				return
+			}
+
+			// Get the response's content type
+			respContentType := resp.Header.Get("Content-Type")
```
nflaig: do all relays properly set this header, otherwise might be best to assume JSON if header is missing

server/get_header.go
```diff
@@ -71,18 +76,60 @@ func (m *BoostService) getHeader(log *logrus.Entry, ua UserAgent, slot phase0.Sl
...
 			if err != nil {
-				log.WithError(err).Warn("error making request to relay")
+				log.WithError(err).Warn("error creating new request")
 				return
 			}
-			if code == http.StatusNoContent {
+
+			// Add headers from the request to this request.
+			// This includes Accept and Eth-Consensus-Version, if provided.
+			for key, values := range header {
```
nflaig: my only worry here would be that relays don't properly handle q-weighted Accept headers, which is fine as long as they default to JSON in that case, if they throw an error it would be problematic though

server/accept.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// AcceptEntry represents a parsed Accept header entry with q-weight.
+type AcceptEntry struct {
+	MediaType string
+	QValue    float64
+}
+
+// ParseAcceptHeader parses and sorts the Accept header by q-values.
+func ParseAcceptHeader(header string) []AcceptEntry {
```
nflaig: if you need some [test cases](https://github.com/ChainSafe/lodestar/blob/unstable/packages/api/test/unit/utils/headers.test.ts#L6)

server/get_header.go
jtraglia: Not sure. Defaulting to JSON sounds like a good idea to me. Will make this change.

server/accept.go
jtraglia: I've incorporated these into this PR. But with two changes. If there is no accept header or the accept header isn't in the supported media types, we will default to JSON.

nflaig: > If there is no accept header or the accept header isn't in the supported media types, we will default to JSON.

we do the same although would return 406 if client only accepts media types we don't support, it's handled [here](https://github.com/ChainSafe/lodestar/blob/f8acf8b56f7c4e96f27e1734868733835153e98d/packages/api/src/utils/server/handler.ts#L42-L57)

jtraglia: Hmm yes. I like that idea. Will do the same.

server/accept.go
```diff
@@ -0,0 +1,134 @@
+package server
```
metachris: Maybe existing battle-tested code for this would have been easier to add: https://github.com/timewasted/go-accept-headers/blob/master/accept.go

server/get_header.go
```diff
@@ -44,12 +52,6 @@ func (m *BoostService) getHeader(log *logrus.Entry, ua UserAgent, slot phase0.Sl
 		"msIntoSlot":  msIntoSlot,
 	}).Infof("getHeader request start - %d milliseconds into slot %d", msIntoSlot, slot)
 
-	// Add request headers
-	headers := map[string]string{
-		HeaderKeySlotUID:      slotUID.String(),
```
metachris, line 49: this gone on purpose?

server/service.go
```diff
@@ -333,6 +342,14 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	m.bids[bidKey(slot, result.bidInfo.blockHash)] = result
 	m.bidsLock.Unlock()
 
+	// How should we respond to the client
```
metachris:
```suggestion
	// Decide response content type (JSON by default)
```

server/service.go
```diff
@@ -343,8 +360,88 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		"relays":      strings.Join(types.RelayEntriesToStrings(result.relays), ", "),
 	}).Info("best bid")
 
-	// Return the bid
-	m.respondOK(w, &result.response)
+	// Define map of supported handlers
+	supportedMediaTypeHandlers := map[string]func(){
+		MediaTypeJSON: func() {
```
metachris: i find this nested pattern confusing, and it makes the whole function complex.

could you break the two out as individual functions? maybe the map also isn't needed, just in the end you request the right function? would simplify things

server/service.go
```diff
@@ -299,29 +300,37 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
 		"pubkey":     pubkey,
 		"ua":         ua,
 	})
-	log.Debug("getHeader")
+	log.Debug("handling request")
+
+	// Additional header fields
+	header := req.Header
+	header.Set("User-Agent", wrapUserAgent(ua))
+	header.Set(HeaderStartTimeUnixMS, fmt.Sprintf("%d", time.Now().UTC().UnixMilli()))
```
metachris: Not a huge fan of doing a bunch of request specific stuff here, feels it should be done inside `m.getHeader` for not leaking and consitency.

server/accept.go
jtraglia: Good call. I only looked for a std library solution. This works great.

server/get_header.go
jtraglia: Nope. My mistake! Fixed.

server/types/relay_entry.go
```diff
@@ -72,3 +73,14 @@ func RelayEntriesToStrings(relays []RelayEntry) []string {
...
 }
+
+// Copy returns a deep copy of the relay entry.
+func (r *RelayEntry) Copy() (ret RelayEntry) {
+	ret.PublicKey = r.PublicKey
+	if r.URL != nil {
+		urlCopy := *r.URL
+		ret.URL = &urlCopy
+	}
+	ret.SupportsSSZ = r.SupportsSSZ
```
metachris: would be a bit cleaner to move it up to line 80, to set all together

server/service.go
```diff
@@ -268,31 +269,36 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		parentHashHex = vars["parent_hash"]
 		pubkey        = vars["pubkey"]
 		ua            = UserAgent(req.Header.Get("User-Agent"))
+		accept        = req.Header.Get("Accept")
```
metachris: perhaps a tad more descriptive name would be useful

```suggestion
		proposerAcceptContentTypes        = req.Header.Get("Accept")
```

server/types/relay_entry.go
jtraglia: Sure. I was just doing in the same order as the struct fields.

server/service.go
```diff
@@ -304,6 +310,14 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	m.bids[bidKey(slot, result.bidInfo.blockHash)] = result
 	m.bidsLock.Unlock()
 
+	// Decide response content type (JSON by default)
+	clientAccepts := goacceptheaders.Parse(req.Header.Get("Accept"))
```
metachris, line 314: since this function deals with many clients and ends, better names would add clarity (i.e. proposer and relay accept headers)

```suggestion
	relayAcceptsContentTypes := goacceptheaders.Parse(req.Header.Get("Accept"))
```

metachris: should be moved up to where proposer accept headers are parsed, at line 272

metachris: Also, should use the `proposerAcceptContentTypes` variable instead of parsing again

server/service.go
```diff
@@ -314,8 +328,40 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
+		preferredContentType = MediaTypeJSON
+	}
+
+	// Respond appropriately
+	if preferredContentType == MediaTypeJSON {
+		m.respondGetHeaderJSON(w, &result)
+	} else if preferredContentType == MediaTypeOctetStream {
+		m.respondGetHeaderSSZ(w, &result)
+	} else {
+		message := fmt.Sprintf("unsupported media type: %s", preferredContentType)
```
metachris: error log here would also be useful

server/service.go
jtraglia: > should be moved up to where proposer accept headers are parsed, at line 272

This feels a little weird. I dislike that we have two very similar variables. I will think about this for a minute.

server/get_header.go
```diff
@@ -189,3 +254,89 @@ func (m *BoostService) getHeader(log *logrus.Entry, ua UserAgent, slot phase0.Sl
...
+}
+
+// respondGetHeaderSSZ responds to the proposer in SSZ
+func (m *BoostService) respondGetHeaderSSZ(w http.ResponseWriter, result *bidResp) {
+	// Serialize the response
+	var err error
+	var sszData []byte
+	switch result.response.Version {
+	case spec.DataVersionBellatrix:
+		w.Header().Set("Eth-Consensus-Version", "bellatrix")
```
metachris: consts for the consensus version would be better than hardcoded strings

server/media_types.go
```diff
@@ -0,0 +1,6 @@
+package server
+
+const (
```
metachris: i would make this file a bit more generic and name it `consts.go`

server/service.go
```diff
@@ -257,35 +258,41 @@ func (m *BoostService) handleRegisterValidator(w http.ResponseWriter, req *http.
...
 func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request) {
 	var (
-		vars          = mux.Vars(req)
-		parentHashHex = vars["parent_hash"]
-		pubkey        = vars["pubkey"]
-		ua            = UserAgent(req.Header.Get("User-Agent"))
+		vars                             = mux.Vars(req)
+		parentHashHex                    = vars["parent_hash"]
+		pubkey                           = vars["pubkey"]
+		ua                               = UserAgent(req.Header.Get("User-Agent"))
```
metachris: a space in between those would help reduce unnecessary spacing

server/service.go
```diff
@@ -307,8 +314,48 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
+	}
+
+	// Get the proposer's highest quality acceptable media type
+	proposerPreferredContentType, err := parsedProposerAcceptContentTypes.Negotiate(MediaTypeJSON, MediaTypeOctetStream)
+	if err != nil {
+		log.WithError(err).Warn("failed to negotiate preferred content-type")
+		proposerPreferredContentType = MediaTypeJSON
+	}
+
+	// If every relay returned the bid in JSON, that means that none
```
metachris: is this logic actually necessary? the bid is decoded already anyway, why not directly return in the proposer preference?

metachris (dismissed):

server/media_types.go
jtraglia: Renamed to `constants.go`.

server/service.go
jtraglia: So if we return SSZ here, the proposer will send SSZ in its getPayload request and then mev-boost will need to convert it back to JSON for each relay. If we return JSON here it saves us the conversion.

It's not strictly necessary. I could remove it if you prefer.

server/get_header.go
jtraglia: I decided to do this for more of these hardcoded strings.

server/service.go
jtraglia: Just pushed a commit to remove this.

metachris (approved):
nice! thanks

PR #735, opened 2025-02-05
Add fixes to SendHTTPRequest

0xTylerHolmes:
## 📝 Summary
1. Adds a missing error check that could lead to a nil deref. 
2. Moves another error check closer to its origin
3. Use similar comment formatting.
## ⛱ Motivation and Context
* (bug) Fixes potential nil-deref by adding a missing error check.
* (nit) Comment formatting
## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


jtraglia (approved):
If there were an error creating the new request, there would indeed be a nil deref in the else block.

PR #742, opened 2025-02-17
Add getPayload SSZ support 

jtraglia:
## 📝 Summary

This PR adds SSZ support for `getPayload` endpoint. Please see:

* https://github.com/ethereum/builder-specs/blob/main/apis/builder/blinded_blocks.yaml

There's also a bit of a refactor away from generics.

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


server/get_payload.go
```diff
@@ -92,35 +124,116 @@ func processPayload[P Payload](m *BoostService, log *logrus.Entry, ua UserAgent,
...
+
+			// Get the resp body content
+			respBytes, err := io.ReadAll(resp.Body)
+			if err != nil {
+				log.WithError(err).Warn("error reading response body")
+				return
+			}
+
+			// Get the response's content type
+			respContentType := resp.Header.Get(HeaderContentType)
```
nflaig: do all relays properly set this header? might be safer to fallback to JSON for now if not set

jtraglia: Good call. I agree we should default to JSON here. Will make this change.

jtraglia: Ah wait. We actually already do this but it's in the `getPayload` function:

https://github.com/flashbots/mev-boost/blob/d7df45be70ffb0228cb4986e7c373efabe4abff8/server/get_payload.go#L49-L54

nflaig: maybe I got confused, thought the code snippet you linked is parsing content type of proposer (ie. CL client calling mev-boost) and the line I commented on is the response from relay/builder to the `submitBlindedBlock` call

jtraglia: Nope. You were right & I was the confused one. Sorry about that.

Fixed here: https://github.com/flashbots/mev-boost/pull/742/commits/5fa6b03fda12a47b3e448243a4a3e38256fb4bb7

server/get_payload.go
```diff
@@ -174,141 +271,318 @@ func verifyPayload[P Payload](payload P, log *logrus.Entry, response *builderApi
...
+				out.Version = spec.DataVersionElectra
+				out.Electra = new(eth2ApiV1Electra.SignedBlindedBeaconBlock)
+				return out.Electra.UnmarshalSSZ(in)
+			default:
+				return errInvalidForkVersion
+			}
+		} else {
+			return types.ErrMissingEthConsensusVersion
+		}
+	case MediaTypeJSON:
```
avalonche, line 430: when decoding json, can ethConsensusVersion be used and have the decoding attempts be a fallback?
I think now all clients will start to add the consensus version as a required field in the header so this can be deprecated when there's no longer a backward compatibility issue

server/service.go
```diff
@@ -342,93 +336,70 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
...
 // handleGetPayload requests the payload from the relays
 func (m *BoostService) handleGetPayload(w http.ResponseWriter, req *http.Request) {
-	log := m.log.WithField("method", "getPayload")
-	log.Debug("getPayload request starts")
+	var (
+		userAgent                        = wrapUserAgent(UserAgent(req.Header.Get(HeaderUserAgent)))
+		proposerContentType              = req.Header.Get(HeaderContentType)
+		proposerAcceptContentTypes       = req.Header.Get(HeaderAccept)
+		parsedProposerAcceptContentTypes = goacceptheaders.Parse(proposerAcceptContentTypes)
+		proposerEthConsensusVersion      = req.Header.Get(HeaderEthConsensusVersion)
```
avalonche, line 346: are there any assumptions here that these headers will be set? are they handled if empty?

jtraglia: They are handled if they're empty. See the following:

`proposerContentType` will default to JSON if it's empty or invalid:

https://github.com/flashbots/mev-boost/blob/94f460400fb7abbdc9e4da79f9cdac0de4703ddd/server/get_payload.go#L50-L54

For accept media types, we will default to JSON too:

https://github.com/flashbots/mev-boost/blob/94f460400fb7abbdc9e4da79f9cdac0de4703ddd/server/service.go#L378-L389

For `proposerEthConsensusVersion`, if the content type is octet-stream and this is empty, we'll return an error.

https://github.com/flashbots/mev-boost/blob/94f460400fb7abbdc9e4da79f9cdac0de4703ddd/server/get_payload.go#L425-L427

server/get_payload.go
jtraglia: This is an interesting idea, but I'm not sure it's worth it. If a client has set `Eth-Consensus-Version`, their preferred content type will almost certainly be SSZ. So I don't think this would ever be used. Also, `Eth-Consensus-Version` [is an optional field](https://github.com/ethereum/builder-specs/blob/1569fc45d43e3ed2bed94ec6788c3f8fa7d2a4b2/apis/builder/header.yaml#L52) and I'm unsure they would set it when sending JSON.

nflaig: > Eth-Consensus-Version [is an optional field](https://github.com/ethereum/builder-specs/blob/1569fc45d43e3ed2bed94ec6788c3f8fa7d2a4b2/apis/builder/header.yaml#L52) and I'm unsure they would set it when sending JSON.

to be clear on the spec side, the intend is to make this mandatory even for JSON similar to how it works on the beacon-api but as we can't expect them to set it right now, it's safer to not rely on it in case of JSON requests / responses

there are some performance improvements that even apply in case of JSON, eg. you can use a predefined schema to parse the data which is generally 2-3x more performant than a general purpose JSON parser

ralexstokes (approved):

PR #751, opened 2025-02-28
update docs/geth-pos-privnet.md

taozui472:
The ganache does not exist in scripts/local_testnet/start_local_testnet.sh. The ganache mentioned here in this file doesn't need to be modified at present.


jtraglia:
That file is pretty out of date. It will require a bigger refactor later. I don't believe removing that item is sufficient.

PR #769, opened 2025-04-18
fix: JSON Content-Type for relay get payload w/o SSZ support

jacobkaufmann:
## 📝 Summary

if the relay does not support SSZ, then use the JSON media type for the `Content-Type` header.

## ⛱ Motivation and Context

in `BoostService.getPayload`, we receive a request from the proposer with a designated `Content-Type`.
that `Content-Type` may either reflect JSON or SSZ (i.e. `octet-stream`).
when we forward the payload request to the relay(s), the relay(s) may not support SSZ.
if a relay does not support SSZ, then we would convert the request payload (i.e. signed blinded block) to JSON.
however, we would not update the `Content-Type` in the request to the relay to reflect the change in the payload.

## 📚 References

N/A

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


jtraglia (approved):
Looks great, thanks for finding this and making a PR!

PR #771, opened 2025-04-21
Update all dependencies

jtraglia:
## 📝 Summary

This PR updates all of the dependencies. Note, I couldn't update the `prysmaticlabs/go-bitfield` package because they moved it but another dependency we use still uses the original package. Will need to wait a bit for that one.

## ✅ I have run these commands

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

PR #778, opened 2025-04-23
Fix parsing of custom genesis time

jtraglia:
## 📝 Summary

I made an incorrect change when updating `urfave/cli` here:
* #771

Apparently, `cmd.Uint64` thinks the value isn't found and returns 0.

Note that this fix is not urgent. It only affects logs when testing with a devnet. Eg this:

https://github.com/flashbots/mev-boost/blob/ee0c91c42fc0d49254498e4d016159fb828c8d8c/server/get_payload.go#L100-L106

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -155,7 +155,7 @@ func setupGenesis(cmd *cli.Command) (string, uint64) {
 	}
 
 	if cmd.IsSet(customGenesisTimeFlag.Name) {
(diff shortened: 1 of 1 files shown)


barnabasbusa (approved):

ralexstokes (approved):
lgtm

metachris (approved):

PR #788, opened 2025-05-01
Send getPayload request to all relays again

jtraglia:
## 📝 Summary

Instead of only sending the getPayload request to relays which provided this bid, send the getPayload request to all relays and convert the signed blinded beacon block to JSON if necessary.

## ⛱ Motivation and Context

In theory, this should improve liveness. If, for example, there is a relay which is geographically closer to the proposer, it might get the getPayload request sooner and therefore it would distribute the block faster.

## 📚 References

https://discord.com/channels/595666850260713488/1341434255154348034/1367203484620947508

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -152,7 +152,7 @@ func (m *BoostService) getPayload(log *logrus.Entry, signedBlindedBeaconBlockByt
 				log.WithField("relaySupportsSSZ", relaySupportsSSZ).Debug("encoding preference")
 
 				// If the relay provided the bid in JSON or did not provide a bid for this payload,
-				// we must convert the signed blinded beacon block to SSZ for this relay
+				// we must convert the signed blinded beacon block from SSZ to JSON for this relay
 				if parsedProposerContentType == MediaTypeOctetStream && !relaySupportsSSZ {
 					requestContentType = MediaTypeJSON
 					startTime := time.Now()
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"slices"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ralexstokes (approved):
i didnt go a fine-grained code review, but agree with the high-level principle here that we query all relays

server/get_payload.go
```diff
@@ -128,41 +129,42 @@ func (m *BoostService) getPayload(log *logrus.Entry, signedBlindedBeaconBlockByt
...
-				// If necessary, use the JSON encoded version and the JSON Content-Type header
+				// Default to the content from the proposer
 				requestContentType := parsedProposerContentType
 				requestBytes := signedBlindedBeaconBlockBytes
-				if parsedProposerContentType == MediaTypeOctetStream && !relay.SupportsSSZ {
-					requestBytes = signedBlindedBeaconBlockBytesJSON
+
+				// Check if the relay supports SSZ
+				relaySupportsSSZ := false
+				for _, originalBidRelay := range originalBid.relays {
```
metachris, line 146: this would be better as a simple map lookup

server/get_payload.go
```diff
@@ -128,41 +129,42 @@ func (m *BoostService) getPayload(log *logrus.Entry, signedBlindedBeaconBlockByt
...
+					if relay.URL == originalBidRelay.URL {
+						relaySupportsSSZ = originalBidRelay.SupportsSSZ
+						break
+					}
+				}
+				log.WithField("relaySupportsSSZ", relaySupportsSSZ).Debug("encoding preference")
+
+				// If the relay provided the bid in JSON or did not provide a bid for this payload,
+				// we must convert the signed blinded beacon block from SSZ to JSON for this relay
+				if parsedProposerContentType == MediaTypeOctetStream && !relaySupportsSSZ {
```
metachris, line 156: not ideal to convert this again for every relay that would need it. would be better to check up-front if any relay will need it, and then prepare it once for all

metachris (approved):
it looks okay, left some comments about possible improvements to the code. def preferable to submit getPayload to all relays.

server/get_payload.go
jtraglia: It cannot be a simple map lookup though. Because the SSZ field might be different.

server/get_payload.go
jtraglia: I disagree. The conversion is quick. Let's just send out the SSZ requests as soon as possible.

OisinKyne:
> to all relays **again**

hey guys, mind if I ask when this changed to the current behaviour? :)

metachris:
@OisinKyne in the latest release (v1.9).

colink14110476:
What's the plan for release normally? Can we get a version of this out soonish?

OisinKyne:
+1 on the above we'd also be keen to see this released in the near term. Maybe a 1.9.1 patch? :)

jtraglia:
Did this change have a noticeable impact on the network?

OisinKyne:
We have seen a noticeable uptick of missed proposals on mev-boost DV clusters, moreso in the last two weeks than since pectra. We have not been able to track down definitively whether its an issue with mime-type headers from the SSZ introduction, certain relays hanging/rate limiting requests, (which causes 950ms by default timeouts that increase the likelihood of orphans), or whether its something related to mev-boost itself. 

One thing we are seeing much more often is the log line:
`level=error msg="no bid for this getPayload payload found, was getHeader called before?"`

While this log might not be related to this PR, and is maybe a function of consensus round timeouts, the behaviour we're hoping for is for mev-boost to send the signedPayload to its relays anyways, whether it has the corresponding bid for it or not. The cluster has signed a header, it won't be signing another (because slashing), so if it could be broadcasted to relays, that would improve liveness in our POV.

server/get_payload.go
metachris: i don't see how it's faster by encoding SSZ for every relay, versus just encoding it once ahead of time. but also fine with me to leave as-is.

metachris:
We'll expedite getting a release out with this now.

PR #805, opened 2025-07-01
Fulu support for mev-boost

bharath-123:
## 📝 Summary

This PR contains the changes required for the upcoming Fulu hard fork.
These changes have been tested locally against rustic-builder using kurtosis

## ⛱ Motivation and Context

We need to still merge the fulu changes in  `github.com/jtraglia/go-builder-client` to main before merging this.

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


common/common.go
```diff
@@ -10,6 +10,16 @@ import (
 
 const (
 	SlotTimeSecMainnet = 12
+	// FieldElementsPerBlob is the number of field elements needed to represent a blob.
+	FieldElementsPerBlob = 4096
+	// BlobExpansionFactor is the factor by which we extend a blob for PeerDas.
+	BlobExpansionFactor = 2
+	// FieldElementsPerCell is the number of field elements in a cell
```
jtraglia:
```suggestion
	// FieldElementsPerCell is the number of field elements in a cell.
```

server/get_payload.go
```diff
@@ -339,31 +340,70 @@ func verifyBlobsBundle(log *logrus.Entry, request *eth2Api.VersionedSignedBlinde
...
 	}
 
+	// Check proofs
+	responseProofs, err := responseBlobsBundle.Proofs()
+	if err != nil {
+		log.WithError(err).Error("failed to get response proofs")
+		return err
+	}
+
+	if request.Version >= spec.DataVersionFulu {
```
jtraglia, line 430: Just a nit, I'd remove this blank line. This would be consistent with the commitments & blobs checks.

common/common.go
```diff
@@ -10,6 +10,16 @@ import (
 
 const (
 	SlotTimeSecMainnet = 12
+	// FieldElementsPerBlob is the number of field elements needed to represent a blob.
+	FieldElementsPerBlob = 4096
+	// BlobExpansionFactor is the factor by which we extend a blob for PeerDas.
```
jtraglia:
```suggestion
	// BlobExpansionFactor is the factor by which we extend a blob for PeerDAS.
```

jtraglia (dismissed):
Thanks @bharath-123! Looks great. Just a few very small nits.

jtraglia:
@metachris is on vacation for the next few weeks.

I will merge this after we get approvals from at least 2 of the following:
* @faheelsattar
* @MariusVanDerWijden
* @ralexstokes

server/get_payload.go
bharath-123: do you mean the blank line in like 373?

faheelsattar (approved):
looks good!

go.mod
```diff
@@ -60,3 +60,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/attestantio/go-builder-client => github.com/jtraglia/go-builder-client v0.4.6-0.20250410195459-42a3ff7f1546
```
MariusVanDerWijden, line 64: Do you want to maintain this for the long term @jtraglia ?

jtraglia: No. We will remove this when attestant adds support. I don't think we have to block merging this PR on that though.

server/get_payload.go
```diff
@@ -339,31 +340,70 @@ func verifyBlobsBundle(log *logrus.Entry, request *eth2Api.VersionedSignedBlinde
...
+				"cellsPerExtBlob":        common.CellsPerExtBlob,
+			}).Error("different lengths for proofs")
+			return errInvalidKZGLength
+		}
+	} else {
+		if len(requestCommitments) != len(responseProofs) {
+			log.WithFields(logrus.Fields{
+				"requestBlobCommitments": len(requestCommitments),
+				"responseProofs":         len(responseProofs),
+			}).Error("different lengths for proofs")
```
MariusVanDerWijden:
```suggestion
			}).Error("wrong lengths for proofs")
```

MariusVanDerWijden: Nitpicky, but I would change all "different" to "wrong"

server/get_payload.go
```diff
@@ -339,31 +340,70 @@ func verifyBlobsBundle(log *logrus.Entry, request *eth2Api.VersionedSignedBlinde
 		return err
 	}
 
-	// Ensure the blobs bundle field counts are correct
-	if len(requestCommitments) != len(responseBlobsBundle.Blobs) ||
-		len(requestCommitments) != len(responseBlobsBundle.Commitments) ||
-		len(requestCommitments) != len(responseBlobsBundle.Proofs) {
+	// Check commitments
+	responseCommitments, err := responseBlobsBundle.Commitments()
```
MariusVanDerWijden, line 400: you could still verify blobs here, would save a bit of duplicated code

server/get_payload.go
```diff
@@ -406,61 +448,78 @@ func convertSSZToJSON(ethConsensusVersion string, sszBytes []byte) ([]byte, erro
...
+			}
+			out.Version = spec.DataVersionCapella
+			out.Capella = block
+		case EthConsensusVersionDeneb:
+			block := new(eth2ApiV1Deneb.SignedBlindedBeaconBlock)
+			if err = json.Unmarshal(in, block); err != nil {
+				return err
+			}
+			out.Version = spec.DataVersionDeneb
+			out.Deneb = block
```
MariusVanDerWijden, line 546: This whole api is so cursed. `out` should just have a type `block any`

server/service_test.go
```diff
@@ -1273,12 +1308,44 @@ func denebExecutionPayloadAndBlobsBundle(header *deneb.ExecutionPayloadHeader, k
 	}
 }
 
+func fuluExecutionPayloadAndBlobsBundle(header *deneb.ExecutionPayloadHeader, kzgCommitments []deneb.KZGCommitment) *builderApiFulu.ExecutionPayloadAndBlobsBundle {
+	numBlobs := len(kzgCommitments)
+	commitments := make([]deneb.KZGCommitment, numBlobs)
+	copy(commitments, kzgCommitments)
+	// For testing, proofs and blobs are not populated
```
MariusVanDerWijden, line 1448: why?

server/service_test.go
```diff
@@ -1273,12 +1308,44 @@ func denebExecutionPayloadAndBlobsBundle(header *deneb.ExecutionPayloadHeader, k
...
+	proofs := make([]deneb.KZGProof, numBlobs)
```
MariusVanDerWijden: Shouldn't this be numBlobs* CellsPerExtBlob

server/get_payload.go
jtraglia: Yes, that's right.

server/get_payload.go
jtraglia: Lol. I don't think making it `any` would be a wise change. It's okay to be explicit.

MariusVanDerWijden: I know, the correct way would be an interface that all the different types implement.
Making it explicit is just super ugly.

edit: just how we implement transactions in geth

server/service_test.go
bharath-123: it would be a good idea to have some default values

jtraglia (dismissed):
LGTM, thanks @bharath-123 & @faheelsattar!

jtraglia (dismissed):

server/service.go
```diff
@@ -337,6 +338,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	}
 }
 
+// Deprecated: For reference: https://github.com/ethereum/builder-specs/issues/119
```
metachris, line 341: nice, thanks for that

server/get_payload.go
```diff
@@ -36,16 +38,39 @@ var (
...
+}
+
+// Deprecated: For reference: https://github.com/ethereum/builder-specs/issues/119
 // getPayload requests the payload (execution payload, blobs bundle, etc) from the relays
 func (m *BoostService) getPayload(log *logrus.Entry, signedBlindedBeaconBlockBytes []byte, userAgent, proposerContentType, proposerAcceptContentTypes, proposerEthConsensusVersion string) (*builderApi.VersionedSubmitBlindedBlockResponse, bidResp) {
+	result, bid := m.innerGetPayload(log, signedBlindedBeaconBlockBytes, userAgent, proposerContentType, proposerAcceptContentTypes, proposerEthConsensusVersion, GetPayloadV1)
+	return result.response, bid
+}
+
+// getPayloadV2 submits the signed blinded beacon block to relays for submission without returning the payload and blobs
```
metachris, line 67: is it necessary to have two functions here? feels like it could be a single function that returns all fields, since the methods themselves are identical

server/get_payload.go
```diff
@@ -111,18 +136,18 @@ func (m *BoostService) getPayload(log *logrus.Entry, signedBlindedBeaconBlockByt
 	originalBid := m.bids[bidKey(slot, blockHash)]
 	m.bidsLock.Unlock()
 	if originalBid.response.IsEmpty() {
-		log.Error("no bid for this getPayload payload found, was getHeader called before?")
+		log.Error("no bid for this payload found, was getHeader called before?")
```
metachris:
```suggestion
		log.Warn("no bid for this payload found, was getHeader called before?")
```

a warn is better here, and it's common for this to happen in DV setups

server/get_payload.go
metachris: ah i see now, the version is passed in

server/get_payload.go
metachris: agree, that'd be nice! perhaps open an issue to remind us of it later and we do a follow-up PR for this?

server/service_test.go
```diff
@@ -1220,20 +1366,41 @@ func blindedBlockToBlockResponse(signedBlock any) *builderApi.VersionedSubmitBli
...
+			panic("failed to convert block")
+		}
 		header := block.Message.Body.ExecutionPayloadHeader
 		commitments := block.Message.Body.BlobKZGCommitments
 		return &builderApi.VersionedSubmitBlindedBlockResponse{
 			Version: spec.DataVersionElectra,
 			Electra: denebExecutionPayloadAndBlobsBundle(header, commitments),
 		}
+	case spec.DataVersionFulu:
+		block, ok := signedBlock.(*eth2ApiV1Electra.SignedBlindedBeaconBlock)
```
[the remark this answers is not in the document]
jtraglia: Because the block didn't change.

metachris (dismissed):

jtraglia (approved):

PR #808, opened 2025-07-11
Update default gas limit as per the latest recommendations

parithosh:
## 📝 Summary

The gas limit recommendation has been updated twice in the last weeks, and this PR is to mirror those updates. A writeup on the testing approach can be found here: https://ethpandaops.io/posts/gaslimit-scaling/

## ⛱ Motivation and Context

We are noticing a lot of validators still signaling 30M on mainnet, this however is contrary to what we hear when speaking to them. We are looking into the datasets to ensure we're gathering the right data, but mev-boost could explain some of the discrepancy. It is possible that when using all defaults, the node is falling back to the default used by mev-boost, this theory is still pending testing - but still worth addressing irrespective of the outcome. 

* [  ] `make lint` (linter already fails from untouched code btw)
* [ x ] `make test-race` 
* [ x ] `go mod tidy`

### DIFF
--- a/cmd/test-cli/README.md
+++ b/cmd/test-cli/README.md
@@ -20,7 +20,7 @@ getPayload [-vd-file] [-mev-boost] [-bn] [-en] [-mm] [-bellatrix-fork-version] [
 
 Env & defaults:
 	[generate]
-	-gas-limit                               = 30000000
+	-gas-limit                               = 45000000
 	-fee-recipient  VALIDATOR_FEE_RECIPIENT  = 0x0000000000000000000000000000000000000000
 
 	[register]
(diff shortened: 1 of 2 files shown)


jtraglia:
Hey @parithosh, this PR will not make a difference. The changes are in the test CLI which is not used.

> mev-boost could explain some of the discrepancy

Not really, mev-boost does not control the gas limit for proposers. The gas limit comes from the consensus layer client via the validator registrations; there is no default value here. And mev-boost [simply forwards these](https://github.com/flashbots/mev-boost/blob/develop/server/register_validator.go) to builders/relays.

> linter already fails from untouched code btw

The checks pass. I suspect you're using newer tool versions.

bharath-123:
ya this PR wouldn't make a lot of difference. But I think it would be still good to change the defaults so that the test-cli by default runs on 45M gas limit.

parithosh:
Yeah thanks for the update! Ill try to track down the 30M discrepancy elsewhere then :D

PR #809, opened 2025-07-21
add support for v2 blinded blocks

faheelsattar:
## 📝 Summary

Introduces v2 blinded blocks endpoint which doesn't return the entire `ExecutionPayload` and `BlobsBundle`

## ⛱ Motivation and Context
Currently the `/eth/v1/builder/blinded_blocks` returns the entire `ExecutionPayload` and `BlobsBundle`. This can increase network bandwidth costs for the proposer and also the request time as we increase the number of Blobs. We therefore introduce a new v2 API endpoint, `/eth/v2/builder/blinded_blocks` which no longer returns the fully signed unblinded payload. Instead, it returns an empty response and relies on the relay to propagate the block.

## 📚 References

https://github.com/ethereum/builder-specs/issues/119
https://github.com/ethereum/builder-specs/pull/123

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -350,7 +350,7 @@ func verifyBlobsBundle(log *logrus.Entry, request *eth2Api.VersionedSignedBlinde
 		log.WithFields(logrus.Fields{
 			"requestBlobCommitments": len(requestCommitments),
 			"responseCommitments":    len(responseCommitments),
-		}).Error("different lengths for commitments")
+		}).Error("wrong lengths for commitments")
 		return errInvalidKZGLength
 	}
 	for i, commitment := range requestCommitments {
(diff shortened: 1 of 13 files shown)


server/get_payload.go
```diff
@@ -44,6 +44,7 @@ var (
 // Core Logic
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// Depreciated: For reference: https://github.com/ethereum/builder-specs/issues/119
```
faheelsattar: haven't added more context here in the code. Let me know if required and we can add some docs here too

server/service.go
```diff
@@ -337,6 +338,7 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 	}
 }
 
+// Depreciated: For reference: https://github.com/ethereum/builder-specs/issues/119
```
faheelsattar: same here

server/get_payload.go
bharath-123:
```suggestion
// Deprecated: For reference: https://github.com/ethereum/builder-specs/issues/119
```

server/service.go
bharath-123:
```suggestion
// Deprecated: For reference: https://github.com/ethereum/builder-specs/issues/119
```

bharath-123: We should definitely add more documentation when we can

server/get_payload.go
bharath-123: A high level description here would be pretty useful for ppl reading the code. You could write a para detailing how it is different from the v1 api and why we need it.

PR #825, opened 2025-09-08
fix: minor security improvements

br3d:
## 📝 Summary

Minor security improvements

## ⛱ Motivation and Context

**CI**:
- Pin actions/checkout to a commit SHA (v4.3.0) across all workflows for supply-chain hardening and reproducible builds.

**Docker**:
- Streamline module copy step and improve build caching.
- Runtime container defaults to root; better to drop privileges.

**Other**
- I also want to draw attention to the following things that I did not change, as there may be flaws in this that I do not know about. 
- The golang image is built on Debian, it may be better to use the image on alpine ?
- Copying /etc/ssl/certs/ca-certificates.crt from the Debian-based builder into Alpine can work, but it’s brittle. It’s safer to install Alpine’s own ca-certificates. 
- Another security recommendation is to vendor all github actions as they may gain access to sensitive information

## 📚 References

[npm debug and chalk packages compromised ](https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised)

---

* [ ] `make lint`
* [ ] `make test-race`
* [ ] `go mod tidy`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 #4.3.0
 
       - name: Set up Go
         uses: actions/setup-go@v3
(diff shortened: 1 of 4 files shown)


metachris:
closing because spam

br3d:
> closing because spam

Are you being serious? Do you fully understand what a supply chain attack is?
Do you understand that running an application from the root user is not a good idea?

PR #686, opened 2024-09-24
feat(issue-253): initial metrics implementation

0xste:
## 📝 Summary
An initial prometheus http server exposed optionally on a custom port if flag is provided at runtime.

## ⛱ Motivation and Context
As per discussion on: https://github.com/flashbots/mev-boost/pull/294#issuecomment-1271171387 i've reduced the scope of the change to be as minimal as possible and look to incrementally add additional metrics over time to de-risk the implementation

Following the step 1 recommendation on: https://github.com/flashbots/mev-boost/issues/253#issuecomment-2371122299

## 📚 References
https://github.com/flashbots/mev-boost/issues/253

---

## ✅ I have run these commands

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

PR #832, opened 2025-09-25
Prometheus Metrics

faheelsattar:
## 📝 Summary
Adds prometheus metrics related to
- bids delivered by relays
- latency by relay
- status of the response received by relays
- status code returned to beacon node
- number of bids below the min bid
- winning bid
- ms into the slot
- last slot for which relay delivered the header

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


cli/flags.go
```diff
@@ -178,4 +183,19 @@ var (
...
+		Usage:    "enables a metrics server",
+		Category: Metrics,
+	}
+	metricsAddrFlag = &cli.StringFlag{
+		Name:     "metrics-addr",
+		Sources:  cli.EnvVars("METRICS_ADDR"),
+		Value:    "localhost:18551",
+		Usage:    "listening address for the metrics server",
+		Category: Metrics,
+	}
```
bharath-123, line 200: minor nit: it would be nice to separate it out to metrics-addr and metrics-port into separate flags.

server/get_header.go
```diff
@@ -94,13 +95,20 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
 
 			// Send the request
 			log.Debug("requesting header")
+			start := time.Now()
 			resp, err := m.httpClientGetHeader.Do(req)
+			if RelayLatency != nil {
+				RelayLatency.WithLabelValues(params.PathGetHeader, relay.String()).Observe(float64(time.Since(start).Milliseconds()))
+			}
```
bharath-123: Let's keep it in microseconds? We can always divide the metric by 1000 to get the value in millisecond. Lets have more precision.

server/get_header.go
```diff
@@ -94,13 +95,20 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
...
 			if err != nil {
 				log.WithError(err).Warn("error calling getHeader on relay")
 				return
 			}
 			defer resp.Body.Close()
 
+			if RelayStatusCode != nil {
+				RelayStatusCode.WithLabelValues(strconv.Itoa(resp.StatusCode), params.PathGetHeader, relay.String()).Inc()
+			}
```
bharath-123: We should keep these behind a function defined in metrics.go. Having a bunch of `if stmts` like this can harm readability

cli/flags.go
jtraglia: This actually mirrors how we do it elsewhere. I'd like to keep it like this for consistency.

https://github.com/flashbots/mev-boost/blob/7e47dd2ab4236157cf42a0373b49d03b95dfdd59/cli/flags.go#L42-L48

jtraglia:
We're missing the README updates from:

* https://github.com/flashbots/mev-boost/pull/686

cli/flags.go
bharath-123: alright sounds good

server/get_header.go
faheelsattar: good suggestion

README.md
```diff
@@ -270,6 +270,10 @@ Usage of mev-boost:
         use Sepolia
   -version
         only print version
+  -metrics
+        enables a metrics server (default: false)
+  -metrics-addr string
+        listening address for the metrics server (default: "localhost:18551")
```
metachris, line 276: are both flags necessary, or should we only use `-metrics-addr` and when set then enable, and by default empty string disable?

jtraglia: Hmm this would require the user provide a URL/port. Not quite as simple as "enable with `-metrics`".

I think we did it this way because it mimics how clients do it, eg:

image

metachris (approved):

PR #839, opened 2025-10-07
Add timing feature for bid optimization

### LINKED ISSUE #830: Add support for timing games features
Allows sending multiple `get_header` requests within a slot and returns the freshest successful header. This helps improving bid while respecting deadlines.

faheelsattar:
## 📝 Summary

Allows sending multiple getHeader requests  within a slot and returns the freshest successful header. This helps improving bid while respecting deadlines. It maximizes chance of a higher bid since later requests can return improved headers. It achieves this via YAML config file which can be passed via `--relay-config` , allowing delayed first requests via `TargetFirstRequestMs` and multiple subsequent requests at intervals via `FrequencyGetHeaderMs`. This feature is only activated if `EnableTimingGames` is set to true in the config file for a particular relay. 

This PR also allows setting relays via YAML file where timing games parameters can be set. This is also backwards compatible if someone doesnt want to set relays via YAML file and want to use the default method of setting relays via `--relay` flag

[example config](https://github.com/faheelsattar/mev-boost/blob/feat/timing-games/config.example.yaml)

```yaml
# Example configuration for mev-boost
# This file can be passed to mev-boost using the -config flag: ./mev-boost -config config.yaml
# the configuration supports hot-reloading, changes to this file will be automatically applied without restarts

# Timeout in milliseconds for get_header requests to relays.
# This value should be less then the timeout on the CL.
# Default: 900ms
timeout_get_header_ms: 900

# Threshold in milliseconds that marks when in a slot is considered "too late".
# If a getHeader request arrives after this threshold, mev-boost skips all relay requests
# and forces local block building.
# Default: 1000ms
late_in_slot_time_ms: 1000

# Relay Configurations
# Each relay can be configured individually. Relays can also be provided via cli using
# the -relay flag. Cli provided relays will be merged with config file relays.

relays:
  # Relay with timing games enabled
  # Timing games allow mev-boost to send multiple requests at strategic intervals
  # to capture the latest, most valuable bids right before the proposal deadline.
  - url: https://0x9000009807ed12c1f08bf4e81c6da3ba8e3fc3d953898ce0102433094e5f22f21102ec057841fcb81978ed1ea0fa8246@relay.relayer1.net
    # Enable timing games strategy for this relay.
    # When enabled, mev-boost will delay the first request and send multiple follow up requests.
    # Default: false
    enable_timing_games: true
    
    # Target time in milliseconds when the first getHeader request should be sent.
    # Only used when enable_timing_games is true.
    # Example: 200 means wait until 200ms before sending the first request.
    target_first_request_ms: 200
    
    # Interval in milliseconds between subsequent getHeader requests to the same relay.
    # After the first request, mev-boost will keep sending new requests every frequency_get_header_ms
    # until the global timeout budget (maxTimeout) is exhausted.
    # Only used when enable_timing_games is true.
    # Example: 100 means send a new request every 100ms.
    frequency_get_header_ms: 100

  # Relay with timing games disabled (standard behavior)
  # This relay will receive a single getHeader request immediately when the CL
  # calls mev-boost, following the standard mev-boost behavior.
  - url: https://0x9000009807ed12c1f08bf4e81c6da3ba8e3fc3d953898ce0102433094e5f22f21102ec057841fcb81978ed1ea0fa8246@relay.relayer2.com
    # Disable timing games - use standard single-request behavior
    enable_timing_games: false
    
    # These values are ignored when enable_timing_games is false
    target_first_request_ms: 0
    frequency_get_header_ms: 0
```
Closes #830 
## ⛱ Motivation and Context
Timing games around bids are a double-edged sword. They can maximize yield for involved parties but also cause more latency (later block proposal within the slot -> more time to include TXs) for the network. Due to sophistication of builders and relays they have already become the norm in the PBS pipeline.
This PR simply provides feature parity with other sidecar implementations and it's still up to proposers whether to opt-in to these advanced features.

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`


examples/relay_config.yml
```diff
@@ -0,0 +1,15 @@
+# Example relay configuration with timing games for mev-boost
+relays:
+  # relay with timing games enabled
+  - url: "https://0x9000009807ed12c1f08bf4e81c6da3ba8e3fc3d953898ce0102433094e5f22f21102ec057841fcb81978ed1ea0fa8246@relay.relayer1.net"
```
jtraglia: For the URLs, let's remove the quotations. These are not required in yaml files.

config.example.yaml
```diff
@@ -0,0 +1,15 @@
...
+    # time in ms from slot start for the first getHeader request
+    target_first_request_ms: 500
+    # time in ms between subsequent getHeader requests
+    frequency_getheader_ms: 100
+
+  # relay with timing games disabled (standard behavior)
+  - url: "https://0x9000009807ed12c1f08bf4e81c6da3ba8e3fc3d953898ce0102433094e5f22f21102ec057841fcb81978ed1ea0fa8246@relay.relayer2.com"
+    enable_timing_games: false
+    target_first_request_ms: 0
+    frequency_getheader_ms: 0
```
jtraglia: It would also work without the bottom three lines right? For example:

```yml
relays:
  - url: https://...
  - url: https://...
  - url: https://...
```

faheelsattar: yeah it would, missing fields will get default values

potuz:
Doesn't this feature require CL client implementation changes to not timeout on the request on the CL side?

faheelsattar:
> Doesn't this feature require CL client implementation changes to not timeout on the request on the CL side?

I think the CL clients sends an optional `X-Timeout-Ms` header which this feature respects.
https://github.com/ethereum/builder-specs/blob/db7c5ee0363c6449997e2a1d5bc9481ed87ba223/apis/builder/header.yaml#L46-#L57

potuz:
CLs have a hardcoded timeout for the return of the header from the call to MeV-Boost, I do not see how this works without modifying the CL

faheelsattar:
> CLs have a hardcoded timeout for the return of the header from the call to MeV-Boost, I do not see how this works without modifying the CL

That is okay too since we also have `TimeoutGetHeaderMs` which will be configurable when setting up mev-boost.
https://github.com/faheelsattar/mev-boost/blob/feat/timing-games/server/get_header.go#L30
https://github.com/faheelsattar/mev-boost/blob/feat/timing-games/server/get_header.go#L85-#L99

potuz:
What I mean is that the CL has a hardcoded timeout that is not configurable.

faheelsattar:
> What I mean is that the CL has a hardcoded timeout that is not configurable.

Yeah i understand what you are saying "the CL clients have hardcoded timeouts (e.g: 1000ms) that isnt configurable". 
This feature is meant to work within that constraint. 
How? cuz we have a configurable value in our mev-boost config file called 'timeout_get_header_ms' (e.g: 900ms) which has to be configured less then the "hardcoded value" in the CL. The operators must configure the values when running mev-boost accordingly.  Hope that makes sense.

potuz:
This can already be (and is) done at the relay level, I still don't see the importance of having this at the middleware level. Anyway hopefully Gloas minimizes this to at most 200-300ms

cli/config.go
```diff
@@ -0,0 +1,104 @@
...
+	"os"
+	"strings"
+
+	"github.com/flashbots/mev-boost/server/types"
+	"gopkg.in/yaml.v3"
+)
+
+type RelayConfigYAML struct {
+	URL                  string `yaml:"url"`
+	ID                   string `yaml:"id"`
```
bharath-123: Do relays need an ID outside the URL?

faheelsattar: optionally if you want. URLs can be too long when being read in metrics so you can essentially give an alias

OisinKyne:
> I think the CL clients sends an optional `X-Timeout-Ms` header which this feature respects.
> [ethereum/builder-specs@`db7c5ee`/apis/builder/header.yaml#L46-#L57](https://github.com/ethereum/builder-specs/blob/db7c5ee0363c6449997e2a1d5bc9481ed87ba223/apis/builder/header.yaml#L46-#L57)

I'm pretty sure this is just on the mev-boost side. I don't think any CLs send this header to the sidecar, nor does the sidecar take the header from that request and re-use it upstream.

cli/config.go
metachris: in other places, just the hostname was a good enough ID and short one. seems better to use only the hostname for metrics too?

metachris:
Why does the ID field in the config exist? 

I think it may not be needed, and MEV-Boost should simply use the full hostname as short version, as it does in other places, not the full URL.

metachris:
@potuz we had multiple conversations with node operators, which all require such a feature to be present to not switch to other implementations that have this (such as commit boost).

potuz:
@metachris I guess it's fine, I don't care much cause hopefully this becomes irrelevant in the next fork, but what I am pointing out is that the relays can (and already do) this, and adding this to mev boost itself is kinda irrelevant in this sense, my reading of this is that mev boost will force a relay that does not do timing games to actually do it. 

I would be worried that adding a delay in the middleware and misscomunication with the relay adding an extra one may cause missed blocks because of breaking the deadline in the CL. I hope these changes get actually tested in deployed environment with real relays and not some mock flashbots builder run by pandaOps.

cli/config.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+
+	var config Config
+	if err := yaml.Unmarshal(data, &config); err != nil {
+		return nil, err
+	}
+
+	timeoutGetHeaderMs := config.TimeoutGetHeaderMs
+	if timeoutGetHeaderMs == 0 {
```
bharath-123, line 144: neeed to clearly specify this in a documentaton on what the default value is. you can do it in the config.yaml

cli/config.go
```diff
@@ -0,0 +1,104 @@
...
+		timeoutGetHeaderMs = 900
+	}
+
+	lateInSlotTimeMs := config.LateInSlotTimeMs
+	if lateInSlotTimeMs == 0 {
+		lateInSlotTimeMs = 1000
```
bharath-123: same for this.

config.example.yaml
```diff
@@ -0,0 +1,24 @@
+# Example configuration for mev-boost
```
bharath-123, line 1: would recommend to create a `config.example.yaml`

cli/config.go
faheelsattar: we already have doc and default values in the config file. Do you propose some different value?

metachris:
Some notes and requests:

1. Document timeout behavior in the README: clearly explain when/why requests start and stop. This is extremely important to clarify well. Maybe could add a [mermaid diagram](https://mermaid.live/edit).
1. Remove `ID` field in config, can use the relay hostname.
1. Hardcoded defaults in `cli/config.go`: `timeoutGetHeaderMs = 900` and `lateInSlotTimeMs = 1000`. What if users actually want to set it to 0? Need to spend a moment to think about that.
1. Add validation for timing parameters: Prevent DoS-level aggressive settings
1. If config file is specified but cannot be watched, it should be an error and fail at startup.
1. Test config hot-reload: Add tests for ConfigWatcher functionality

#### Error handling when config file cannot be watched

Location: [cli/config.go:82-107](https://github.com/faheelsattar/mev-boost/blob/020d85bd970be1c3b33d17c15e487dafa540fec1/cli/config.go#L82-L108) and [cli/main.go:115](https://github.com/faheelsattar/mev-boost/blob/020d85bd970be1c3b33d17c15e487dafa540fec1/cli/main.go#L115-L116)

Issue: Config watcher failures are logged as warnings but should fail at startup:

```
if err != nil {
    log.WithError(err).Warn("failed to set up config watcher")
}
```

#### Clearly document

For example (should be more detailed): 
- Delays first request until `TargetFirstRequestMs` if configured
- Sends repeated requests at `FrequencyGetHeaderMs` intervals
- Stops when maxTimeoutMs is reached (calculated from `timeoutGetHeaderMs` and `lateInSlotTimeMs`)
- Returns the latest (freshest) bid with highest value

README.md
```diff
@@ -317,6 +317,80 @@ Example for setting a minimum bid value of 0.06 ETH:
...
+
+```bash
+./mev-boost -config config.yaml
+```
+**Notice:** This feature is strictly meant for advanced users and extra care should be taken when setting up timing game associated parameters
+
+#### 1. Global Timeouts
+These settings apply to all relays and define the hard boundaries for the `getHeader` operation.
+
+*   **`timeout_get_header_ms`** (optional, default: 900ms)
```
canercidam: How did we choose defaults 900 and 1000? And could we suggest a general practice to the users for picking different numbers?

config.example.yaml
```diff
@@ -0,0 +1,51 @@
...
```
canercidam, line 1: By introducing a config file like this, we introduce to users two files to manage (the second is the one that contains the mev-boost run command). Usually it is hard for users to understand what happens when config values and flag values are combined like this. And it is easy to forget old values in either of these places after e.g. attempting to remove a relay entirely when it is in fact added to both places. So it could be less confusing if we did one of these in addition:
- If a config file is provided, let all relays be configured in the config file and exit with an error if any relay is provided from the flag.
- Allow relays from config file and flag values but disallow mentioning one relay in both places (can be checked in the merge func).

cli/config.go
```diff
@@ -0,0 +1,173 @@
...
+	return &ConfigWatcher{
+		v:          v,
+		configPath: absPath,
+		cliRelays:  cliRelays,
+		log:        log,
+	}, nil
+}
+
+// Watch starts watching the config file for changes
+func (cw *ConfigWatcher) Watch(onConfigChange func(*ConfigResult)) {
```
canercidam, line 86: We seem to introduce a new watcher along with the config file, which is very nice and is super useful when experimenting with the timing games logic quickly. But when enabled by default, this can cause unwanted side effects since the user might save a few times while making changes. I think I'd want to make some changes and then do a restart with the new values, in production. Maybe we could introduce a simple `-watch-config` flag to enable the behavior when it is more helpful to use?

README.md
```diff
@@ -317,6 +317,80 @@ Example for setting a minimum bid value of 0.06 ETH:
 
 Optionally, the `-metrics` flag can be provided to expose a prometheus metrics server. The metrics server address/port can be changed with the `-metrics-addr` (e.g., `-metrics-addr localhost:9009`) flag.
 
+### Enabling timing games
```
canercidam: Thanks a lot for this documentation! :slightly_smiling_face: Can we keep on mentioning "Enable timing games" here but below the "Notice:" sentence, share a link to the rest of the details (config, explanations, diagram etc.). The link could point to a new file `docs/timing-games.md` and the main README would look less bloated without the advanced feature details.

server/get_header.go
```diff
@@ -59,218 +70,384 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
...
+	m.relayConfigsLock.RLock()
+	relayConfigs := m.relayConfigs
+	timeoutGetHeaderMs := m.timeoutGetHeaderMs
+	lateInSlotTimeMs := m.lateInSlotTimeMs
+	m.relayConfigsLock.RUnlock()
 
-		// Relays that sent the bid for a specific blockHash
-		relays = make(map[BlockHashHex][]types.RelayEntry)
-	)
+	if timeoutGetHeaderMs < lateInSlotTimeMs-msIntoSlot {
```
canercidam, line 107: It would be really great to add a comment line that explains the timeout calculation logic briefly. It's better when it's easier to reason about a piece of logic like this when reading through. :slightly_smiling_face:

server/get_header.go
```diff
@@ -59,218 +70,384 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
...
+		}
+	}
 
-			// Ignore bids with an empty block
-			if bidInfo.blockHash == nilHash {
-				log.Warn("relay responded with empty block hash")
-				return
-			}
+	// send multiple requests at frequency intervals
+	if relayConfig.FrequencyGetHeaderMs > 0 { //nolint:nestif
```
canercidam: Without a positive frequency value, there is no “timing games” logic, right? Just a one-off request delayed by the first request target parameter. Maybe we should require this frequency value as non-zero if timing games are enabled and not leave room for a weird case? (The config parser could enforce that validation and let the process throw an error accordingly.)

Also, do you remember what the issue was with the linter? Maybe it’s better to fix the “nestif” related problem instead of disabling the linter.

server/get_header.go
```diff
@@ -59,218 +70,384 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
...
+			"frequencyMs":   relayConfig.FrequencyGetHeaderMs,
+			"timeoutLeftMs": timeoutLeftMs,
+		}).Debug("sending multiple header requests via timing games")
+
+		var bidResults []bidResult
+		var mu sync.Mutex
+
+		// keep sending requests until time runs out
+		var wg sync.WaitGroup
+		for timeoutLeftMs > 0 {
```
canercidam: Did we consider using `time.Ticker` with `select` inside this `for` to launch `go func()`? We could do something like:
```go
timeoutCh := time.After(timeoutLeftMs)
ticker := time.NewTicker(time.Duration(relayConfig.FrequencyGetHeaderMs) * time.Millisecond)
for {
	select {
		case <-ticker.C:
			// go func()
		case <-timeoutCh:
			break
	}
}
```

server/get_header.go
```diff
@@ -59,218 +70,384 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
...
+		log.WithFields(logrus.Fields{
+			"frequencyMs":   relayConfig.FrequencyGetHeaderMs,
+			"timeoutLeftMs": timeoutLeftMs,
+		}).Debug("sending multiple header requests via timing games")
+
+		var bidResults []bidResult
+		var mu sync.Mutex
+
+		// keep sending requests until time runs out
+		var wg sync.WaitGroup
```
canercidam: It is nice to use wg here but it would be great to ensure that the HTTP request has some timeout. Otherwise, this wg will never be completely done if the outgoing request hangs. The alternative to wg use could be making requests synchronously, never having competing requests launched against one relay from here and avoiding the extra goroutines.

(In other words, since we're adding more concurrency, it would be great to do something to reduce the chances of leaking more goroutines. :slightly_smiling_face:)

canercidam (commented):
Great stuff @faheelsattar! Noting down some improvements in mind. Could you check the tests, too? They seem to be failing. Thanks! :slightly_smiling_face:

config.example.yaml
canercidam: (Resolved)

cli/config.go
canercidam: (Resolved)

server/get_header.go
canercidam: (Resolved)

server/get_header.go
canercidam: We discussed keeping the logic so it's fine. But still need to explore the linter problem.

server/get_header.go
canercidam: (Resolved)

server/get_header.go
canercidam: (Resolved)

README.md
canercidam: - The new default for `timeout_get_header_ms` is 950 based on a past discussion.
- The new default for `late_in_slot_time_ms` is 2000 because I guess it's a more reliable default value for production.

The reasoning is a little vague but we apparently tried to pick the best numbers we can so we can consider this as resolved.

server/get_header.go
faheelsattar: restructured our conditions to fix the lint

README.md
faheelsattar: yepp, make sense, added improved docs

canercidam (approved):

metachris (approved):

PR #840, opened 2025-10-08
feat: fallback to v1 getPayload api if v2 getPayload api fails

bharath-123:
## 📝 Summary

We want to fallback calling the v1 getPayload API if the v2 getPayload API fails. 

https://github.com/ethereum/builder-specs/pull/123 introduced a new v2 getPayload API for the relays. This essentially doesn't return the entire signed beacon block response which can really large with more blobs. 

If in case, relays do not implement this API. The getPayload v2 API will return a 404 not found and the slot will be missed. It would be better to fallback to the v1 API in such a case to avoid a missed slot.

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -236,9 +236,9 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 				// Check that the response was successful
 
 				// If the relay does not support V2 API, retry with V1 API
-				// we can fallback to V1 API if the status code returned >= 400. There is no harm 
-				// falling back to the V1 API, falling back to the V1 API in the case of any error 
-				// can be beneficial to the proposer to avoid a missed slot. 
+				// we can fallback to V1 API if the status code returned >= 400. There is no harm
+				// falling back to the V1 API, falling back to the V1 API in the case of any error
+				// can be beneficial to the proposer to avoid a missed slot.
 				if resp.StatusCode >= http.StatusBadRequest && url == relay.GetURI(params.PathGetPayloadV2) {
 					log.Warn("relay may not support V2 API, Retrying with V1 API")
 					// retry with v1 api
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -235,8 +235,12 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 				RecordRelayStatusCode(strconv.Itoa(statusCode), endpoint, relay.String())
 				// Check that the response was successful
 
-				if resp.StatusCode == http.StatusNotFound && url == relay.GetURI(params.PathGetPayloadV2) {
-					log.Warn("relay may not support V2 API, Retrying with v1 API")
+				// If the relay does not support V2 API, retry with V1 API
+				// we can fallback to V1 API if the status code returned >= 400. There is no harm 
+				// falling back to the V1 API, falling back to the V1 API in the case of any error 
+				// can be beneficial to the proposer to avoid a missed slot. 
+				if resp.StatusCode >= http.StatusBadRequest && url == relay.GetURI(params.PathGetPayloadV2) {
+					log.Warn("relay may not support V2 API, Retrying with V1 API")
 					// retry with v1 api
 					url = relay.GetURI(params.PathGetPayload)
 					versionToUse = GetPayloadV1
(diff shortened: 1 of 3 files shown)


server/get_payload.go
```diff
@@ -225,12 +231,20 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
...
 
 				RecordRelayStatusCode(strconv.Itoa(statusCode), endpoint, relay.String())
 				// Check that the response was successful
+
+				if resp.StatusCode == http.StatusNotFound && url == relay.GetURI(params.PathGetPayloadV2) {
+					log.Warn("relay may not support V2 API, Retrying with v1 API")
+					// retry with v1 api
+					url = relay.GetURI(params.PathGetPayload)
+					versionToUse = GetPayloadV1
+					return nil, fmt.Errorf("relay may not support V2 API, Retrying with v1 API")
```
bharath-123: We leverage the existing retry loop to call the v1 API in the upcoming retry.

server/get_payload.go
```diff
@@ -248,7 +262,7 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 			var result payloadResult
 			result.success = true
 
-			if version == GetPayloadV1 {
+			if versionToUse == GetPayloadV1 {
```
bharath-123, line 266: We should definitely be validating the returned block by the relay even if we fall back to the v1 API.

server/service.go
```diff
@@ -33,6 +33,7 @@ var (
 	errInvalidPubkey             = errors.New("invalid pubkey")
 	errNoSuccessfulRelayResponse = errors.New("no successful relay response")
 	errServerAlreadyRunning      = errors.New("server already running")
+	errRetryWithV1API            = errors.New("relay may not support V2 API, retrying with v1 API")
```
faheelsattar: "V1"

server/get_payload.go
```diff
@@ -157,14 +157,15 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 
 	for _, relay := range m.relays {
 		go func(relay types.RelayEntry) {
+			var versionToUse GetPayloadVersion
```
faheelsattar: we can reuse `version` and update it where needed, but i think `versionToUse` sounds more mutable since now it can also be updated. Why dont we rename `version` to `versionToUse` in the `innerGetPayload` args and then use it everywhere

bharath-123: sounds good let me do that

server/get_payload.go
```diff
@@ -225,12 +228,20 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
...
 				if err != nil {
-					log.WithError(err).Warnf("error calling getPayload%s on relay", version)
+					log.WithError(err).Warnf("error calling getPayload%s on relay", versionToUse)
 					return nil, err
 				}
 
 				RecordRelayStatusCode(strconv.Itoa(statusCode), endpoint, relay.String())
 				// Check that the response was successful
+
+				if resp.StatusCode == http.StatusNotFound && url == relay.GetURI(params.PathGetPayloadV2) {
```
metachris: maybe more robust to do this on any 4xx status code, instead of only not found

server/get_payload.go
```diff
@@ -225,12 +228,20 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
...
+					log.Warn("relay may not support V2 API, Retrying with v1 API")
+					// retry with v1 api
+					url = relay.GetURI(params.PathGetPayload)
+					versionToUse = GetPayloadV1
+					return nil, errRetryWithV1API
```
metachris, line 247: where is `errRetryWithV1API` used? where is the actual retry with v1 triggered and executed?

server/get_payload.go
bharath-123: that is not a bad idea. There is no harm falling back to the v1 api eitherways.

server/get_payload.go
bharath-123: `errRetryWithV1API` is not used anywhere. so this logic of calling the API is encapsulated in the `retry` method which retries if the function passed to it returns an error. `errRetryWithV1API` is an error that is returned when the V2 API is not found or so.

server/get_payload.go
bharath-123: we could just fallback if the status code >= 400 in general i believe

metachris (approved):

PR #850, opened 2025-11-04
allow background relay requests in getPayloadV2

faheelsattar:
## 📝 Summary

This PR allows sending `getPayloadV2` endpoint to all relays by allowing relay requests to continue in the background after receiving the first successful response.

This PR also involves  improving the logging around MEV-Boost:
1. We add color to the logs to improve readability
2. in the GetPayload request handle, we ensure that we log the url and the errorBody upon response if there is an error.

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -166,6 +166,8 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 
 			// If the request fails, try again a few times with 100ms between tries
 			resp, err := retry(requestCtx, m.requestMaxRetries, 100*time.Millisecond, func() (*http.Response, error) {
+				log = log.WithField("url", url)
+
 				// Default to the content from the proposer
 				requestContentType := parsedProposerContentType
 				requestBytes := signedBlindedBeaconBlockBytes
@@ -240,6 +242,7 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 				// falling back to the V1 API, falling back to the V1 API in the case of any error
 				// can be beneficial to the proposer to avoid a missed slot.
 				if resp.StatusCode >= http.StatusBadRequest && url == relay.GetURI(params.PathGetPayloadV2) {
+					log.WithError(err).Warn("unexpected status code")
 					log.Warn("relay may not support V2 API, Retrying with V1 API")
 					// retry with v1 api
 					url = relay.GetURI(params.PathGetPayload)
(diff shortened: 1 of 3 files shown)


server/get_payload.go
```diff
@@ -154,6 +154,7 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 	// Create a context with a timeout as configured in the http client
 	requestCtx, requestCtxCancel := context.WithTimeout(context.Background(), m.httpClientGetPayload.Timeout)
 	defer requestCtxCancel()
+	originalVersionToUse := versionToUse
```
bharath-123: don't use an `originalVersionToUse` here. Each relay request is a separate go-routine and we need to handle each separately. 
I think we just don't call `requestCtxCancel` at this point and let the other requests run.

server/get_payload.go
```diff
@@ -235,20 +236,29 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
...
+					}
+					errorBodyStr := string(errorBody)
+					log.WithField("errorBody", errorBodyStr).Warnf("unexpected status code %d", resp.StatusCode)
+
+					// If the relay does not support V2 API, retry with V1 API
+					// we can fallback to V1 API if the status code returned >= 400. There is no harm
+					// falling back to the V1 API, falling back to the V1 API in the case of any error
+					// can be beneficial to the proposer to avoid a missed slot.
+					if resp.StatusCode >= http.StatusBadRequest && url == relay.GetURI(params.PathGetPayloadV2) {
+						log.Warn("relay may not support V2 API, Retrying with V1 API")
```
jtraglia: The capitalized "Retrying" word should be lowercase.
```suggestion
						log.Warn("relay may not support V2 endpoint, retrying with V1 endpoint")
```

jtraglia: Or this if you prefer:
```suggestion
						log.Warn("relay may not support getPayloadV2, retrying with getPayloadV1")
```

server/get_payload.go
```diff
@@ -235,20 +236,29 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
...
+						// retry with v1 api
```
jtraglia, line 255: Just a nit, you say "V1 API" (uppercase) in the comments above. A bit consistent.

jtraglia (approved):
LGTM, but maybe fix the little warning message nit.

PR #859, opened 2025-11-11
chore: Tag logs with the relay URLs in getPayload

bharath-123:
## 📝 Summary

Tag logs with relay URLs in getPayload for better readability. 

we call getPayload parallely across multiple relays and it can get hard to identify which log is coming from which relay invocation. Tagging the logs by URLs can help us understand the logs better!

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/server/get_payload.go
+++ b/server/get_payload.go
@@ -167,6 +167,7 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 
 			// If the request fails, try again a few times with 100ms between tries
 			resp, err := retry(requestCtx, m.requestMaxRetries, 100*time.Millisecond, func() (*http.Response, error) {
+				innerLog = innerLog.WithField("url", url)
 				// Default to the content from the proposer
 				requestContentType := parsedProposerContentType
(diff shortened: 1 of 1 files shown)


server/get_payload.go
```diff
@@ -166,8 +166,6 @@ func (m *BoostService) innerGetPayload(log *logrus.Entry, signedBlindedBeaconBlo
 
 			// If the request fails, try again a few times with 100ms between tries
 			resp, err := retry(requestCtx, m.requestMaxRetries, 100*time.Millisecond, func() (*http.Response, error) {
-				log = log.WithField("url", url)
```
bharath-123, line 169: This was not right since it can cause a race condition with other getPayload invocations.

metachris: Wouldn't it be better to create a new logger instance there, instead of adding url to every single log call?

ie. `_log = log.WithField...`

bharath-123: The log is a pointer and each go-routine would override that pointer. this can cause a race conditions amongst multiple getPayload calls where we see the wrong url for the wrong invocation.

jtraglia (approved):

PR #871, opened 2025-11-30
add color flag for text log

### LINKED ISSUE #867: logrus formatted changed to instantiate with ForceColors true
In #850, specifically in 59da2e8481867685240e6bcbb3f19a7d6e11e2d6, the logrus formatter was updated to force colored output.

This change not only enables colors, but also changes the formatting of log level and timestamps. 

I'm requesting to add support for non-colored standard `logfmt` support, just as JSON formatting is supported (flag `-json`).

I have no preference for how this feature is implemented. 
I see a number of ways to accomplish this. A few examples:

- add flag `-log-format=json|color|logfmt` while retaining deprecated support for `-json`
- add flag `-color`, thereby reverting the force color change
- add flag `-no-color` to support pre-1.10.1 formatting

Alternatively, a simple fix could involve:

- removing the `ForceColors: true` change from 59da2e8481867685240e6bcbb3f19a7d6e11e2d6

Let me know what you prefer and I'll submit a PR

faheelsattar:
## 📝 Summary

Introduces an optional `color` flag for text log fmt
Closes #867

## ⛱ Motivation and Context

Previously we were enforcing color for text logs, this change makes it configurable, so users can choose depending on their preference.  

## 📚 References

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ Usage of mev-boost:
   -json
         log in JSON format instead of text
   -color
-        enable colored output for text log format
+        enable colored output for text log format; has no effect if JSON logging is enabled via -json
   -log-no-version
         disables adding the version to every log entry
   -log-service string
--- a/README.md
+++ b/README.md
@@ -252,6 +252,8 @@ Usage of mev-boost:
         use Hoodi
   -json
         log in JSON format instead of text
+  -color
+        enable colored output for text log format
   -log-no-version
         disables adding the version to every log entry
   -log-service string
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -252,6 +252,8 @@ Usage of mev-boost:
         use Hoodi
   -json
         log in JSON format instead of text
+  -color
```
bharath-123, line 255: good to document that, this won't have any effect if `-json` is selected. A small sentence would be helpful. 

Something like:
`This option won't have any effect if -json is enabled.`

metachris (approved):

bharath-123 (approved):

Issue #489, opened 2023-04-08
Info msg "no bid received" misleading for bids below min-bid


KgNoop:
When calling the relays:
// Call the relays
	var mu sync.Mutex
	var wg sync.WaitGroup
	for _, relay := range m.relays

the answers of the relays get checked for various circumstances. One of which is the check against the min-bid:

// Skip if value (fee) is lower than the minimum bid
			if responsePayload.Value().Cmp(m.relayMinBid.BigInt()) == -1 {
				log.Debug("ignoring bid below min-bid value")
				return}

If the bid is below the min bid it does not get added to the Hash of Bids. 
If all set relays are below min bid the Has is empty and the result/Info given bei the logs is "no bid received" which is misleading and makes room for questions about the setup.

if result.blockHash == "" {
		log.Info("no bid received")
		w.WriteHeader(http.StatusNoContent)
		return}

Would be better to have an additional/different msg/info if the relays "all bids below min-bid - creating own block"

metachris:
yeah that'd be nice

PR #892, opened 2026-02-19
server: track bids received to clarify empty response logs

### LINKED ISSUE #869: "no bid received" info level log is misleading
At least myself and [one other person](https://www.reddit.com/r/ethstaker/comments/12enr2i/mevboost_msgno_bid_received/) have been a bit mislead during the debugging process of trying to figure out why a block was built locally. 

In the event that no bids were received that exceed the min-bid setting, with log level set at the default (above debug), the user will see the info level log "no bid received" that is triggered in handleGetHeader if result.response.isEmpty is true. 

In reality, bids may have actually been received, but the "bid received" and the potential subsequent "ignoring bid below min-bid value" logs from inside getHeader are at debug level, and won't show up at the default log level above debug.

The info level log from handleGetHeader if result.response.isEmpty is true should better reflect the real meaning of having result.response.isEmpty, which might be something like "no acceptable bids received" or something like that.

mark-liu:
## 📝 Summary

When all relay bids are filtered (e.g. below min-bid), the info log says
"no bid received" — but bids were received, they just weren't acceptable.
This tracks the count of bids returned by relays in `bidResp` and uses it
to distinguish between genuinely receiving nothing vs filtering everything out.

- `bidsReceived` set from `len(relayBids)` after relay goroutines complete
- Log says `"no acceptable bid received" bidsReceived=N` when bids were filtered
- Log says `"no bid received"` when relays truly returned nothing
- Two new subtests in `TestGetHeaderBids` covering both cases

## ⛱ Motivation and Context

Fixes #869. Also addresses #489.

Running at default (info) log level, the only signal a validator gets when
bids exist but fall below min-bid is "no bid received". The debug-level
"bid received" and "ignoring bid below min-bid value" logs don't show up
unless you explicitly lower the level. This has confused operators debugging
why blocks were built locally.

## 📚 References

- #869
- #489

---

* [x] `make lint`
* [x] `make test-race`
* [x] `go mod tidy`

Note: `make test-race` times out on `develop` (without this PR's changes)
due to `TestGetPayload/"A relay which does not provide the bid gets a JSON
request"` hanging under the race detector. All tests prior to that pass,
including the new tests added here.

### DIFF
--- a/server/get_header.go
+++ b/server/get_header.go
@@ -158,7 +158,7 @@ func (m *BoostService) getHeader(log *logrus.Entry, slot phase0.Slot, pubkey, pa
 	wg.Wait()
 
 	var (
-		result = bidResp{}
+		result = bidResp{bidsReceived: len(relayBids)}
 		relays = make(map[BlockHashHex][]types.RelayEntry)
 	)
 
--- a/server/service.go
+++ b/server/service.go
@@ -334,9 +334,13 @@ func (m *BoostService) handleGetHeader(w http.ResponseWriter, req *http.Request)
 		return
 	}
 
-	// Bail if none of the relays returned a bid
+	// Bail if none of the relays returned an acceptable bid
 	if result.response.IsEmpty() {
-		log.Info("no bid received")
+		if result.bidsReceived > 0 {
+			log.WithField("bidsReceived", result.bidsReceived).Info("no acceptable bid received")
+		} else {
+			log.Info("no bid received")
+		}
 		IncrementBeaconNodeStatus(strconv.Itoa(http.StatusNoContent), params.PathGetHeader)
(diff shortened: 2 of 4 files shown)


mark-liu:
Gentle ping — this is a small observability improvement. Happy to adjust if needed.

mark-liu:
Friendly ping — adds bid tracking so the "no bids received" log message distinguishes between no bids vs bids filtered out. Happy to address feedback.

mark-liu:
Friendly ping — this tracks bid counts to distinguish "no bids received" from "bids received but none selected" in relay responses. Happy to address feedback.

mark-liu:
Closing — no engagement after 90+ days and 2 pings. Happy to reopen if anyone wants to revisit clearer empty-response logging.

PR #895, opened 2026-03-04
fix JSON log marshaling error in registerValidator

### LINKED ISSUE #894: bug: JSON log format breaks on registerValidator due to *http.Request field containing func() type
### Description

When running mev-boost with JSON logging (`-json` flag), the following error is printed to stderr on every `registerValidator` call:

```
Failed to obtain reader, failed to marshal fields to JSON, json: unsupported type: func() (io.ReadCloser, error)
```

### Root Cause

In `server/register_validator.go:50-52`, the full `*http.Request` struct is passed as a logrus field:

```go
log.WithFields(logrus.Fields{
    "request": req,
}).Debug("sending the registerValidator request")
```

`*http.Request` contains the field `GetBody func() (io.ReadCloser, error)`, which Go's `encoding/json` cannot marshal. When logrus's `JSONFormatter` calls `json.Marshal()` on the log fields, it fails with:

```
json: unsupported type: func() (io.ReadCloser, error)
```

Logrus (v1.9.3, used by mev-boost) catches this in `entry.write()` and prints the error to stderr. The log entry is then **silently dropped**.

This works fine with the text formatter (which uses `fmt.Sprintf("%v", req)`), but breaks with `-json`.

### Affected Versions

Confirmed in **v1.11** and **v1.12** (same code present in both).

dom-nie:
## Summary

- Remove `*http.Request` from logrus fields in `registerValidator` debug log
- The `http.Request.GetBody` field (`func() (io.ReadCloser, error)`) causes `json.Marshal` to fail when running with `-json` log formatting, printing `Failed to obtain reader` to stderr and silently dropping the log entry
- The relay `url` is already present on the logger context (set at line 35), so no useful information is lost

Closes #894

### DIFF
--- a/server/register_validator.go
+++ b/server/register_validator.go
@@ -47,9 +47,7 @@ func (m *BoostService) registerValidator(log *logrus.Entry, regBytes []byte, hea
 				req.Header[key] = values
 			}
 
-			log.WithFields(logrus.Fields{
-				"request": req,
-			}).Debug("sending the registerValidator request")
+			log.Debug("sending the registerValidator request")
 
 			// Send the request
 			start := time.Now()


0xsarawut:
> ## Summary
> * Remove `*http.Request` from logrus fields in `registerValidator` debug log
> * The `http.Request.GetBody` field (`func() (io.ReadCloser, error)`) causes `json.Marshal` to fail when running with `-json` log formatting, printing `Failed to obtain reader` to stderr and silently dropping the log entry
> * The relay `url` is already present on the logger context (set at line 35), so no useful information is lost
> 
> Closes #894

aitrading_979814402_virtual@swdq4q31noemail.com

PR #896, opened 2026-03-20
fix: replace *http.Request in log field with JSON-safe fields

### LINKED ISSUE #894: bug: JSON log format breaks on registerValidator due to *http.Request field containing func() type

seeforschauer:
## Change Description

Replace raw `*http.Request` struct in logrus field with individual JSON-serializable fields (`method`, `url`, `contentType`).

`*http.Request` contains `GetBody func() (io.ReadCloser, error)` which `encoding/json` cannot marshal. When mev-boost runs with `-json` flag, logrus `JSONFormatter` fails to serialize the field and **silently drops the log entry**.

## Root Cause

`server/register_validator.go:51` passed the full request struct:
```go
log.WithFields(logrus.Fields{
    "request": req,  // contains func() — breaks json.Marshal
}).Debug(...)
```

## Fix

Extract the useful fields instead:
```go
log.WithFields(logrus.Fields{
    "method":      req.Method,
    "url":         req.URL.String(),
    "contentType": req.Header.Get("Content-Type"),
}).Debug(...)
```

Fixes #894.

## Testing

- `go build ./...` — passes
- `go test ./server/...` — all pass

### DIFF
--- a/server/register_validator.go
+++ b/server/register_validator.go
@@ -48,7 +48,9 @@ func (m *BoostService) registerValidator(log *logrus.Entry, regBytes []byte, hea
 			}
 
 			log.WithFields(logrus.Fields{
-				"request": req,
+				"method":      req.Method,
+				"url":         req.URL.String(),
+				"contentType": req.Header.Get("Content-Type"),
 			}).Debug("sending the registerValidator request")
 
 			// Send the request


seeforschauer:
Closing — duplicate of #895 by @dom-nie which addresses the same issue. Apologies for not checking existing PRs first.

